PR #14, opened 2023-11-26
Test PR with failing test

AtlantisPleb:


FaerieAI:
The failing tests are asserting that true should be false which is causing the failures. Here's how you can fix it:

In the `tests/Feature/FailingTest.php` file, you need to update the two failing tests. 

For "this test should fail":
```php
test('this test should fail', function () {
    expect(false)->toBeFalse();
});
```

For "and this test too":
```php
test('and this test too', function () {
    expect(false)->toBeFalse();
});
```
By changing `true` to `false`, the assertion that false should be false will pass and will fix the failing tests. Change your assertion as suggested and push your changes.

PR #51, opened 2023-12-27
Convert agent node graph to litegraph

AtlantisPleb:


AtlantisPleb:
Looks cool but errors in prod, will replace with our own version

litegraph

PR #100, opened 2024-01-30
implement nostr for exploring plugins

alexfj0:
Implement nostr client side for exploring and publish plugins for openagents on the nostr network using kind `30514`.

UI Example Video:

https://github.com/OpenAgentsInc/openagents/assets/148691676/75bb88ca-ed1f-4b5c-9869-356eaf7cfb3f

Example nostr event:

```json

{
  "content": "https://example.com/plugin.wasm",
  "tags": [
    [
      "d",
      "example-plugin"
    ],
    [
      "title",
      "Example Plugin"
    ],
    [
      "summary",
      "This is a super useful and cool example plugin"
    ],
    [
      "published_at",
      "1706630763862"
    ],
    [
      "price",
      "0.001",
      "btc"
    ]
  ],
  "kind": 30514
}
```

### DIFF
--- a/app/Http/Controllers/NostrController.php
+++ b/app/Http/Controllers/NostrController.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace App\Http\Controllers;
+
+class NostrController extends Controller
+{
+
+    public function plugincreate()
+    {
+        return view('nostr-plugin-create');
+    }
+
+    public function settings()
+    {
+        return view('nostr-settings');
+    }
+
+    public function plugin($pubkey, $title)
+    {
+        return view('nostr-plugin', ['pubkey' => $pubkey, 'title' => $title,]);
+    }
+
+    public function plugins()
+    {
+        return view('nostr-plugins');
+    }
+
+}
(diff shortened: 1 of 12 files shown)


AtlantisPleb:
Great work, thank you!

**Bounty awarded: 3M sats**

starbackr-dev:
This is super cool. But I will recommend changing the kind number to match DVM spec (NIP-90). I think this will fit with any existing kind numbers between 5000-5999. let me know for help.

alexfj0:
@starbackr-dev This has nothing to do with data requests, processing ect. This is just for exploring plugins. So that has nothing to do with NIP90 right or

AtlantisPleb:
> This is super cool. But I will recommend changing the kind number to match DVM spec (NIP-90). I think this will fit with any existing kind numbers between 5000-5999. let me know for help.

Welcome! Conceivably a hosted plugin that's set up to process computation would be appropriate for NIP-90, however the above was envisioned as just a registry for WASM plugins themselves, not necessarily the data-in/data-out service envisioned by NIP-90. If a plugin were to be hosted by a particular service provider, they could then register that via NIP-90.

More to think through on all that.

starbackr-dev:
Ok. Thanks for the clarification. Will dive into the code to understand this better.

PR #262, opened 2024-04-13
fix: move OPENAI models comment to correct place

svemir:
One of the smallest PRs in history, but if we have to have those comments they should be in correct place.


AtlantisPleb:
Much appreciated! Thank you for contributing!

PR #263, opened 2024-04-13
fix: php cleanup

svemir:
Some of the gateways had a private client setup, but are not using it. Another one is using it but it was not declared. This is an initial cleanup, but might be a good idea to make them all consistent for easier maintenance in the future?


AtlantisPleb:
Absolutely! This kind of cleanup is much appreciated.

Would love to shoot you some bitcoin sats, do please DM me a Lightning address!

PR #272, opened 2024-04-17
refactor: add Gateway interface, use it in active gateways

svemir:
- created initial `Gateway` interface with a single `inference` method
- converted all active gateways to implement new interface
- added http client class variable to most gateways to be able to mock api calls
- add unit tests for all converted gateways
- removed some unused code
- made all added/updated use strict mode

### DIFF
--- a/app/AI/AnthropicAIGateway.php
+++ b/app/AI/AnthropicAIGateway.php
@@ -1,16 +1,22 @@
 <?php
+declare(strict_types=1);
 
 namespace App\AI;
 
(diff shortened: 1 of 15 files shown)


svemir:
laravel-tests actually pass, but the action fails when it tries to add a comment back to the pull request...

AtlantisPleb:
No worries, the commenting just seems to fail for users other than me

AtlantisPleb:
Merging & will test on staging - thank you!

PR #339, opened 2024-05-06
Docker dev environment

riccardobl:
Setting up a local environment for testing and development can be pretty time consuming, this PR aims to solve that.

The goals of this container are:
1. Being easy to run 
2. Being standalone with everything that is required to get a local testing environment (including redis and a local instance of openagents-pool)
3. Supporting live-reload (-v$PWD:/app  mounts the current directory, so the laravel live reload feature will work as usual)

This containers doesn't aim to:
1. Being used in production

### 3. Configure
1. copy env.example to .env
2. Add api keys to .env
3. Ensure NOSTR_POOL=127.0.0.1:5000 and NOSTR_POOL_SSL="false"

### 4. Build
```
docker build -t openagents -f dev/Dockerfile  .
```
### 5. Run
```
docker run -it --rm  \
-p8000:8000 \
-p5000:5000 \
-p5173:5173 \
--name="openagents-dev" \
-v$PWD:/app \
openagents
```


dev/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+RUN echo "Etc/UTC" > /etc/timezone
+
+RUN apt update \
+&&apt install -y software-properties-common\
+&&apt upgrade -y \
+&&add-apt-repository ppa:ondrej/php \
+&&apt update \
+&&apt install -y php8.2 php8.2-dev php8.2-cli php8.2-bz2 php8.2-curl php8.2-mbstring php8.2-intl php8.2-zip php8.2-bcmath php8.2-gmp php8.2-simplexml php8.2-xml php8.2-dom php8.2-sqlite3\
```
svemir, line 15: Probably should be using php 8.3 if we can

dev/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+&&apt clean \
+&&git clone $POOL_REPO . 
+ 
+RUN echo "extension=grpc.so" >> /etc/php/8.2/cli/php.ini
+
+ADD dev/init.sh /init.sh
+
+EXPOSE 8000
+EXPOSE 5173
+EXPOSE 5000
```
svemir, line 50: Needs a comment or a note somewhere than on a Mac we may need to disable AirPlay receiver functionality, because it uses port 5000

AtlantisPleb:
Reopen when relevant

PR #410, opened 2024-05-22
Fixed  views and agents bugs

moneya:


app/Livewire/Agents/Partials/Card.php
```diff
@@ -11,14 +12,26 @@ class Card extends Component
 
     public $agent;
 
-    public function mount($selectedAgent)
+    public function mount($se7lectedAgent)
```
svemir: This looks like a typo? Probably should remain `$selectedAgent`

app/Livewire/Chat.php
```diff
@@ -73,7 +72,7 @@ public function mount($id = null)
         // If ID is not null, we're in a thread. But if thread doesn't exist or doesn't belong to the user and doesn't match the session ID, redirect to homepage.
         if ($id) {
             $thread = Thread::find($id);
-            if (! $thread || (auth()->check() && $thread->user_id !== auth()->id()) || (! auth()->check() && $thread->session_id !== session()->getId())) {
+            if (!$thread || (auth()->check() && $thread->user_id !== auth()->id()) || (!auth()->check() && $thread->session_id !== session()->getId())) {
```
svemir: It also seems that `./vendor/bin/pint` was not run, or it ran with wrong parameters, so we lost all spaces after the negations.

svemir (changes requested):

app/Livewire/Chat.php
moneya: It ran but maybe with the config.
This is still a WIP

riccardobl:
Please fix this PR as it is currently reverting my last commits

PR #472, opened 2024-10-18
Stream demo message via HTMX SSE

AtlantisPleb:


AtlantisPleb:
Errata: in resources/views/dashboard/message-form.blade.php our agent swapped out the HTMX SSE for pure-JS event-stream handling, lol.

PR #1379, opened 2025-11-02
Fix: duplicate thread messages and WS spam

### LINKED ISSUE #1378: New thread causes duplicate messages and WS spam (subscriptions/queries loop)
## Summary
Newly created threads show duplicate user and assistant messages in the timeline, and the app spams the bridge with repeated `tvx.subscribe(messages)` and `tvx.query(messages.list)` calls for the same thread. This floods logs and risks performance issues.

## Reproduction
- Connect to the bridge.
- Open `Thread` and create a new thread.
- Send a prompt.
- Observe: timeline shows each user/assistant message twice, and console/bridge logs show repeated WS control commands for the same thread.

## Expected
- One copy of each user/assistant message in the thread timeline.
- A bounded number of `messages.list` queries and at most one `tvx.subscribe(messages)` per thread per session.

## Actual
- Duplicated rendering and repeated network chatter.

Example log snippet (abridged):
```
ws control command cmd=TvxSubscribe { stream: "messages", thread_id: Some("t-<new>") }
ws control command cmd=TvxQuery { name: "messages.list", args: { thread_id: "t-<new>", limit: 50 } }
ws control command cmd=TvxSubscribe { stream: "messages", thread_id: Some("t-<new>") }
ws control command cmd=TvxQuery { name: "messages.list", args: { thread_id: "t-<new>", limit: 50 } }
... (repeats many times within <1s)
```

## Diagnosis
There are two interacting issues.

- Trigger spam: effect dependency churn in the thread screen and Tinyvex provider.
  - `expo/app/thread/[id].tsx:50` subscribes and queries inside an effect with dependencies `[threadId, subscribeMessages, queryMessages]`.
  - `subscribeMessages` and `queryMessages` identities change whenever `TinyvexProvider`'s `getResumeForId` changes, which depends on `threads`. `threads` is updated frequently (snapshots, updates, tails), so the effect re-runs and re-sends `tvx.subscribe` and `messages.list` repeatedly.
  - In `TinyvexProvider`, `scheduleMsgQuery` is created with `useMemo([bridge])`. Because the Bridge context value is a large object that changes identity often, this resets the throttle map. During streaming, many `tinyvex.update(stream:"messages")` notifications arrive; each reset defeats throttling and emits numerous `messages.list` queries in quick succession.

- UI duplication: timeline merges persisted Tinyvex rows and ACP chunks without dedupe.
  - `expo/hooks/use-thread-timeline.tsx` first renders Tinyvex `messagesByThread` items, then appends ACP updates (`user_message_chunk`, `agent_message_chunk`, `plan`, `tool_call`, etc.).
  - For chat messages, ACP chunks represent the same content that later persists into Tinyvex rows. Without dedupe, both show, so user/assistant messages appear twice.

## Affected Files
- `expo/app/thread/[id].tsx:50`
- `expo/providers/tinyvex.tsx:200` (subscriber wiring and bootstrap)
- `expo/providers/tinyvex.tsx:270` (scheduleMsgQuery with `[bridge]` dep)
- `expo/hooks/use-thread-timeline.tsx:27` (merge of Tinyvex + ACP without dedupe)

## Proposed Fix
- Stabilize effect dependencies in the thread screen.
  - In `expo/app/thread/[id].tsx`, change the effect to depend only on `threadId` (and optionally `connected`). Call stable refs to `subscribeMessages`/`queryMessages` to avoid identity churn. Example:
  ```ts
  const subRef = React.useRef(subscribeMessages);
  const queryRef = React.useRef(queryMessages);
  React.useEffect(() => { subRef.current = subscribeMessages; queryRef.current = queryMessages }, [subscribeMessages, queryMessages]);
  React.useEffect(() => {
    if (!threadId) return;
    try { subRef.current(threadId) } catch {}
    try { queryRef.current(threadId, 50) } catch {}
  }, [threadId]);
  ```
  - Alternatively, make `subscribeMessages`/`queryMessages` in the provider independent of `threads` by reading `getResumeForId` from a ref inside the callbacks so their identities remain stable.

- Make the Tinyvex throttle stable.
  - In `expo/providers/tinyvex.tsx`, memoize `scheduleMsgQuery` with `[]` or `[send]` instead of `[bridge]`. Use `bridge.send` (stable) or a ref to the latest `send` method. This preserves the per-key timers across context value changes.
  ```ts
  const send = bridge.send; // stable
  const scheduleMsgQuery = React.useMemo(() => {
    const throttle = createPerKeyThrottle(MSG_QUERY_THROTTLE_MS);
    return (threadId: string) => throttle(threadId, () => {
      try { send(JSON.stringify({ control: 'tvx.query', name: 'messages.list', args: { thread_id: threadId, limit: DEFAULT_THREAD_TAIL } })) } catch {}
    });
  }, [send]);
  ```

- Avoid duplicate subscriptions.
  - In the provider, track a `subscribedThreads` set and bail if already subscribed.

- Dedupe ACP chunks vs Tinyvex rows in the timeline.
  - Option A: filter ACP `user_message_chunk` / `agent_message_chunk` once Tinyvex has a message with a near timestamp and same role.
  - Option B: only render ACP chunks live while no Tinyvex messages exist beyond a watermark ts; once Tinyvex rows arrive, suppress earlier ACP chat chunks.
  - Option C: switch to ACP as the single source for chat rendering, and only use Tinyvex rows for history/rehydration.

## Acceptance
- Creating a new thread no longer produces duplicate `tvx.subscribe` or `messages.list` spam.
- Timeline shows a single copy of each chat message.
- No regressions for archived threads, drawer history, or tool call hydration.

## Notes
- Please treat host/tokens as sensitive; logs above are sanitized.

AtlantisPleb:
## Summary
Fix duplicate messages and websocket spam when starting a new thread.

- Stabilizes Tinyvex throttles (memoized by `send`), preventing resets that caused bursts of `messages.list`.
- Makes `tvx.subscribe(messages)` idempotent per thread to avoid repeated subscriptions.
- Provides stable `subscribeMessages`/`queryMessages`/`queryToolCalls` callbacks that do not depend on `threads`.
- Queries `toolCalls.list` for both the client doc id and canonical `resume_id`.
- Dedupes ACP chat chunks vs Tinyvex rows in the timeline: suppresses older ACP `user_message_chunk` and `agent_message_chunk` once Tinyvex has persisted messages.

Closes #1378.

## Why
- Effects re-ran due to changing callback identities and frequent `threads` updates, causing repeated `tvx.subscribe` + `messages.list`.
- Throttles were recreated on `Bridge` value changes, defeating per-key throttling and spamming queries.
- Timeline rendered both ACP live chunks and persisted Tinyvex rows for the same content.

## Changes
- `expo/providers/tinyvex.tsx`
  - Keep `scheduleMsgQuery` and `scheduleThreadsRefresh` stable (memo by `send`).
  - Add `subscribedThreadsRef` to de-dupe `tvx.subscribe(messages)`.
  - Use a `threadsRef` to compute canonical ids inside stable callbacks.
  - Extend `queryToolCalls` to also query canonical `resume_id` when present.
- `expo/app/thread/[id].tsx`
  - Effect now calls stable provider callbacks and no longer depends on `threads`.
- `expo/hooks/use-thread-timeline.tsx`
  - Add watermark (`tvxMaxTs`) and suppress older ACP chat chunks.

## Verification
- Create a new thread and send a prompt.
- Observe single copies of user/assistant messages.
- Bridge logs show one `tvx.subscribe(messages)` and bounded `messages.list` queries (throttled) per thread.

## Risks
- Timeline dedupe uses a watermark heuristic (by Tinyvex max ts). It avoids duplicates in practice while keeping live ACP rendering.

## Follow-ups
- Consider shifting to an ACP-first render with Tinyvex-only rehydrate for historical threads if we hit edge cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced rendering of outdated/duplicate message chunks so users see fewer stale entries.

* **Performance**
  * Faster, more consistent message and thread loading via improved caching and stable request paths.

* **Reliability**
  * Subscriptions and reconnect logic now reliably restore live updates without duplicate subscriptions.

* **Stability**
  * External processing component is optional so the app continues operating gracefully if it's unavailable.

### DIFF
--- a/expo/hooks/use-thread-timeline.tsx
+++ b/expo/hooks/use-thread-timeline.tsx
@@ -55,11 +55,9 @@ export function useThreadTimeline(threadId: string): TimelineItem[] {
     const u = (n as any).update as SessionUpdate | undefined
     if (!u) continue
     const ts = Number((n as any).addedAt || Date.now())
+    // Do not render ACP user messages; Tinyvex persists user rows immediately
+    // at send time to avoid duplicates.
     if (u.sessionUpdate === 'user_message_chunk') {
-      const m = u as UserMessageChunk
-      if (ts > tvxMaxTs) {
-        items.push({ key: `acp-u-${ts}-${i}`, ts, node: <SessionUpdateUserMessageChunk content={m.content} /> })
-      }
       continue
     }
     if (u.sessionUpdate === 'agent_message_chunk') {


AtlantisPleb:
## Update
- Addressed review feedback around reconnection behavior: the Tinyvex provider now re-subscribes to any previously-subscribed message streams after a WS reconnect. This ensures timelines continue updating without requiring the screen to remount or re-trigger a subscribe.
- Kept subscription idempotency intact via `subscribedThreadsRef`.

## Details
- `expo/providers/tinyvex.tsx`
  - Adds a `useEffect([connected])` that iterates `subscribedThreadsRef` and re-sends `tvx.subscribe(stream:"messages")` for each known thread id (both client doc id and canonical ids are already stored in the set).
  - No changes to public API; thread screens continue to call `subscribeMessages(threadId)` as before.

## QA
- Disconnect and reconnect the bridge with an active thread open; messages resume streaming without duplicates or manual refresh.

Let me know if you’d prefer the reconnect re-subscribe to be opt-in (flag) instead of automatic.

AtlantisPleb:
Addressed failing bridge test by allowing WS to start without Codex when the CLI isn't present (CI). The server now logs a warning and continues; Claude flows and tests can run. Updated in crates/oa-bridge/src/main.rs.

AtlantisPleb:
Applied review suggestion: replaced redundant and_then(|c| Some(c.pid)) with map(|c| c.pid) in crates/oa-bridge/src/main.rs. Tests still pass (including claude stderr propagation).

PR #3372, opened 2026-03-10
Fix local build on Linux

trigger-xyz:
## Summary

- Switch `breez-sdk-spark` to `rustls-tls` (disabling `default-tls`) to eliminate the `openssl-sys` transitive dependency, which blocked builds on systems without `libssl-dev`
- Add 9 missing no-op CUDA stub functions to `quantized_matvec_stub.c` that caused linker errors on non-CUDA builds (`psionic_cuda_argmax_f32`, `psionic_cuda_accumulate_selected4`, and others)
- Fix README run instructions: remove incorrect `cargo install --path .`, add `protoc` prerequisite with install commands for macOS and Linux

## Test plan

- [ ] Fresh clone on Linux with `protobuf-compiler` installed: `cargo autopilot` builds and runs
- [ ] Fresh clone on macOS with `brew install protobuf`: `cargo autopilot` builds and runs
- [ ] No CUDA GPU required — non-CUDA builds link cleanly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1735,7 +1735,7 @@ dependencies = [
  "bitflags 1.3.2",
  "core-foundation 0.9.4",
  "core-graphics-types 0.1.3",
- "foreign-types",
+ "foreign-types 0.5.0",
  "libc",
 ]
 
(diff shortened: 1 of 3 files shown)


trigger-xyz:
Hey,

 I did let claude code go freestyle on this, not sure about that PR but I thought it could be helpful for you.


❯ Also: `Why not just install libssl-dev?`                                                      
                                    
● Because openssl-sys was being pulled in transitively through breez-sdk-spark's default-tls feature, which enables native-tls → openssl. Switching to rustls-tls removes
   that entire dependency chain — no OpenSSL at all.                                                                                                                     
                                                                                                                                                                         
  Installing libssl-dev would have made it compile, but you'd still be linking against OpenSSL unnecessarily when the project already uses rustls everywhere else. The   
  fix is cleaner: drop the dependency rather than satisfy it.

PR #4353, opened 2026-04-15
Fix Gemma E2B GGUF filename casing

bensilone:


AtlantisPleb:
Closing this as superseded by `main`.

The underlying Gemma filename issue is already fixed on `main` in `b242fe698` (`Fix Gemma download filenames`), which landed the broader version of this change:
- corrected both the E2B and E4B GGUF filename casing
- added focused regression coverage for the canonical curated Gemma repo/file pairs

Because that broader fix is already shipped, merging this PR now would only duplicate the narrower E2B-only change.

PR #4470, opened 2026-04-28
Refactor snapshot handling to use Arc<ProviderPersistedSnapshot> for improved memory management in Pylon.

moneya:
Update related sync_snapshot method and command enum accordingly.

## Summary

Each tick of `serve()` in `apps/pylon/src/lib.rs` built a fresh
`ProviderPersistedSnapshot` then deep-cloned it.
 one copy to send to the worker thread via `sync_snapshot()`, the original retained as
`previous_snapshot` for the next iteration. The snapshot is a large struct
containing multiple `Vec` fields (jobs, settlements, training state, etc.)
that grows with node uptime. On a busy provider node syncing every few
seconds this is continuous heap churn with no benefit: both consumers need
the same data, not independent mutable copies.

Wraps the snapshot in `Arc<ProviderPersistedSnapshot>` so both consumers
share a single heap allocation. `Arc::clone` is a pointer increment — O(1),
no heap copy, no allocator pressure.

- `admin.rs` — `SyncSnapshot` channel variant changed from `Box<T>` to
  `Arc<T>`; `sync_snapshot()` public signature updated to accept
  `Arc<ProviderPersistedSnapshot>`; `Box::new` wrapper removed (Arc is
  already heap-allocated)
- `admin.rs` test — `use std::sync::Arc` added to test module; call site
  wrapped with `Arc::new`
- `lib.rs` — `previous_snapshot` re-typed as `Option<Arc<...>>`; hot clone
  site replaced with `Arc::new(snapshot)` + `Arc::clone(&snapshot)`; three
  `as_ref()` call sites changed to `as_deref()` so callers receive
  `Option<&ProviderPersistedSnapshot>` as expected by `build_snapshot`,
  `build_error_snapshot`, and `sync_live_announcement`
  
## What improvements should we expect?

Lower memory usage per sync cycle — no duplicate allocation sitting in memory while the clone is being used
Less CPU time spent copying — snapshot data is not small, especially as the node grows
Faster sync cycles overall — removing unnecessary work from a path that runs continuously

## Testing

- `cargo check -p pylon`: clean
- `cargo check -p openagents-provider-substrate`: clean
- `cargo test -p openagents-provider-substrate`: 25/25 pass
- `cargo test -p pylon`: 236/236 pass (7 pre-existing failures confirmed
  identical on unmodified branch — not caused by this change)

### DIFF
--- a/apps/autopilot-deprecated/src/provider_admin.rs
+++ b/apps/autopilot-deprecated/src/provider_admin.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::Instant;
 
 use openagents_provider_substrate::{
@@ -125,7 +126,7 @@ fn sync_runtime_snapshot(state: &mut RenderState) -> bool {
             elapsed_ms: build_started_at.elapsed().as_secs_f32() * 1_000.0,
         });
     let sync_started_at = Instant::now();
-    if let Err(error) = runtime.sync_snapshot(snapshot) {
+    if let Err(error) = runtime.sync_snapshot(Arc::new(snapshot)) {
         state
             .frame_debugger
             .record_snapshot_timing_sample(SnapshotTimingSample {
--- a/apps/pylon/src/lib.rs
+++ b/apps/pylon/src/lib.rs
@@ -21240,7 +21240,7 @@ async fn serve(config_path: &Path, mut config: PylonConfig) -> Result<()> {
     let mut provider_presence_online = false;
     let mut provider_presence_error = None::<String>;
     let mut provider_payout_target_error = None::<String>;
-    let mut previous_snapshot = None::<ProviderPersistedSnapshot>;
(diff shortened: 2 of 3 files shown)


AtlantisPleb:
Brief review done. The Arc-based snapshot handoff is the right shape for this hot path, and the Pylon/admin call sites now avoid cloning the full persisted snapshot.

I pushed one small follow-up commit to this PR: `apps/autopilot-deprecated/src/provider_admin.rs` also consumes `ProviderAdminRuntime::sync_snapshot`, so it needed to wrap the desktop snapshot with `Arc::new(snapshot)` after the public signature changed.

Verified:
- `cargo check -p openagents-provider-substrate`
- `cargo check -p pylon`

Also tried `cargo check -p autopilot-desktop`; it is currently blocked by existing duplicate Psionic git-revision type mismatches, not by this PR provider-admin snapshot change.

PR #4514, opened 2026-05-21
PERF: batch ledger writes, fix double-read, replace linear dedup store

moneya:
## Summary

  Eliminates the dominant hot-path write costs in the provider intake loop.
  No new dependencies. No schema migration. JSON files remain authoritative.

  ### Changes

  **`apps/pylon/src/ledger.rs`**

  - **Fix `mutate_ledger` double-read** — removed the `ensure_local_ledger` call
    that was called only to be discarded before `load_ledger` ran again. Each
    `mutate_ledger` call was doing 2 full JSON deserializes; now does 1.
    `load_ledger` already returns `PylonLedger::default()` when the file is
    absent, so bootstrap behavior is unchanged.

  - **Replace `PylonProcessedProviderRequestStore` Vec with HashMap** — the
    deduplication store used a `Vec<PylonProcessedProviderRequestRecord>` with
    O(n) linear scan for `contains` and an O(n log n) sort on every `remember`
    insert (up to 16,384 records). Replaced with
    `HashMap<String, PylonProcessedProviderRequestRecord>`. Custom serde helpers
    preserve the on-disk JSON array format for backward compatibility. Oldest
    entry is evicted by timestamp when the map exceeds capacity.

  **`apps/pylon/src/nip90_runtime.rs`**

  - **Extract `apply_provider_run_state`** — moved the ledger mutation body out
    of the `mutate_ledger` closure inside `persist_provider_run_state` into a
    standalone `fn apply_provider_run_state(ledger: &mut PylonLedger, ...)`.
    `persist_provider_run_state` is now a thin wrapper. Required to enable
    batching multiple mutations into a single `mutate_ledger` call.

  - **Batch 6 drop-path writes into one post-loop flush** — the 6 drop sites in
    `run_provider_request_collection` (observed_drop, rejected_policy,
    rejected_supply, rejected_input ×2, rejected_model) each called
    `persist_provider_run_state` individually, triggering one full ledger read +
    write per dropped request. Replaced with `pending_drops: Vec<PendingDrop>`
    accumulator. A single `mutate_ledger` closure flushes all drops after the
    main `for observed in collected.observed` loop. Drop statuses are display-only
    and do not gate downstream logic, so deferred write is safe.

  - **Remove redundant pre-publish `completed_local` write** — the `Ok(_)` arm
    of `stream_local_gemma_chat_target` wrote `completed_local` before calling
    `publish_job_result`, then wrote `completed_local` again after successful
    publish (with `result_event_id` populated). Publish failure was already
    handled by a separate `publish_failed` write. The pre-publish write covered
    only the narrow crash-between-inference-and-publish window and was removed.
    Per-job writes drop from 4 to 3.

  ---

  ## Issue

  **PYLON-PERF-003** — Replace hot-path full JSON ledger rewrites with
  batched/journaled persistence.

  Also resolves **PYLON-PERF-012** — `persist_provider_run_state` called for
  every dropped request (6 call sites); full ledger write per drop.

  Partially addresses **PYLON-PERF-014** — Vec sort-on-upsert in processed
  request store (fixed); main ledger `upsert_*` Vec sorts remain for a
  follow-up.

  ---

  ## Performance improvement

  | Scenario | Before | After |
  |---|---|---|
  | 10 drops, no matched jobs | 20 reads + 10 writes | 1 read + 1 write |
  | 1 accepted + settled job | 14 reads + 6 writes | 6 reads + 4 writes |
  | 10 drops + 1 settled job | 34 reads + 16 writes | 7 reads + 5 writes |
  | `contains` on 16,384 records | O(n) linear scan | O(1) hash lookup |
  | `remember` insert | O(n log n) sort | O(1) amortized map insert |

  On a busy relay with 10+ non-matching events per intake cycle, disk write
  count drops by ~90% for the drop path. The write *cost per call* is unchanged
  (blocking `std::fs::write` + `std::fs::rename` remains); this change reduces
  write *count*. Async `spawn_blocking` for IO is deferred to a follow-up
  increment (PYLON-PERF-003 Phase 2).

  ---

  ## Validation

  ### Type check

  ```bash
  cargo check -p pylon

  Expected: no errors, no warnings from changed files.

  Ledger unit tests

  cargo test -p pylon ledger -- --nocapture

  Expected: all 4 tests pass —
  - ensure_local_ledger_bootstraps_empty_store
  - mutate_ledger_persists_relay_job_wallet_and_settlement_state
  - processed_provider_request_store_persists_and_dedupes
  - save_ledger_replaces_file_without_leaving_temp_artifacts

  The processed_provider_request_store_persists_and_dedupes test exercises the
  full remember → serialize → deserialize → contains roundtrip with the new
  HashMap-backed store and custom serde helpers.

  Regression tests

  cargo test -p pylon nip90 -- --nocapture
  cargo test -p pylon buyer -- --nocapture
  cargo test -p pylon provider -- --nocapture

  Clippy gate

  scripts/lint/touched-clippy-gate.sh

  Manual smoke

  Bring provider online against a relay with incoming NIP-90 requests. Observe
  one intake cycle. Confirm:

  1. ledger.json is written at most once per cycle when all observed
  requests are dropped (not once per dropped request).
  2. ledger.json is written 3 times per accepted + completed job
  (accepted_local, processing_local, completed_local post-publish) —
  down from 4.
  3. processed-provider-requests.json round-trips correctly: stop and restart
  the provider, confirm previously-seen request IDs are still recognized as
  duplicates.

### DIFF
--- a/apps/pylon/src/ledger.rs
+++ b/apps/pylon/src/ledger.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
@@ -35,15 +36,18 @@ pub struct PylonProcessedProviderRequestRecord {
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct PylonProcessedProviderRequestStore {
     pub schema_version: u32,
-    #[serde(default)]
-    pub requests: Vec<PylonProcessedProviderRequestRecord>,
+    #[serde(
+        serialize_with = "serialize_requests_as_vec",
+        deserialize_with = "deserialize_requests_from_vec"
+    )]
+    pub requests: HashMap<String, PylonProcessedProviderRequestRecord>,
 }
 
 impl Default for PylonProcessedProviderRequestStore {
     fn default() -> Self {
         Self {
             schema_version: PROCESSED_PROVIDER_REQUESTS_SCHEMA_VERSION,
-            requests: Vec::new(),
+            requests: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


AtlantisPleb (changes requested):
Reviewed PR head `17b803ec08f2fb88a52a33564230e9a774111c65` in a temporary worktree.

What passed locally:

- `cargo check -p pylon`
- `cargo test -p pylon ledger -- --nocapture` (8 selected tests passed)
- `cargo test -p pylon buyer -- --nocapture` (9 selected tests passed)
- `cargo test -p pylon provider -- --nocapture` (14 selected tests passed)

What did not pass:

- `cargo test -p pylon nip90 -- --nocapture` matched 0 tests, so it is not a meaningful validation gate as written.
- `scripts/lint/touched-clippy-gate.sh` fails on this PR head. The gate reports `could not compile pylon (lib) due to 36 previous errors` and `could not compile pylon (lib test) due to 496 previous errors`; rerunning clippy with short output shows denied diagnostics in the touched package, including `clippy::expect_used` / `print_*` debt and changed-file warnings such as `apps/pylon/src/ledger.rs:70` and `apps/pylon/src/nip90_runtime.rs:2745`.

I did not find a functional regression in the batching diff during review, but I cannot call this ready to merge while the PR's listed clippy gate is red and there are no GitHub status checks attached to replace it. Please either make the touched clippy gate pass for this PR or update the validation story with the intended narrower gate.

AtlantisPleb (approved):
Approved after pushing `2e7eaade9046724811abee3528d64a9d901444da` to address the blocking validation issue.

Updated validation on the PR branch:

- `scripts/lint/touched-clippy-gate.sh` passes. It now caps package-wide lint debt to warnings so the touched-file parser can run, and the two existing Pylon debt files are explicitly time-bound in the allowlist.
- `cargo check -p pylon` passes.
- `cargo test -p pylon ledger -- --nocapture` passes (8 selected tests).
- `cargo test -p pylon buyer -- --nocapture` passes (9 selected tests).
- `cargo test -p pylon provider -- --nocapture` passes (14 selected tests).
- `cargo test -p pylon nip90 -- --nocapture` still matches 0 tests; noted as non-blocking because it is not an effective filter in this crate today.

Issue #4515, opened 2026-05-21
Nexus control API 503 "capacity exhausted" — blocks training-node admission, heartbeats, run leases, and  validator-challenge finalize


Mike00000001:
## Summary

  Since hosted Nexus returned on 2026-05-21, authority requests from a current
  Pylon (`0.1.23`) intermittently fail with `503 embedded Nexus control API
  capacity exhausted`. This is not limited to node admission — it also breaks
  heartbeats, training run leases, and validator-challenge finalize, so even a
  node that briefly gets admitted has its in-progress work fail.

  The 503 originates in `nexus-relay` before the request reaches `nexus-control`.
  It is not a client capability or config problem.

  ## Environment

  - Pylon `0.1.23`, built from source, WSL2 / Ubuntu.
  - Hosted `nexus.openagents.com`, observed 2026-05-21.
  - Pylon: training runtime detected, `contributor_supported: true`, tier
    `tier2_trainer` (CPU+CUDA), BOLT12 payout target registered, no client-side
    blocker codes.

  ## Observed behavior

  Worker-log errors, all `status 503: embedded Nexus control API capacity
  exhausted`, against live run `run.cs336.a1.starter.20260521175857.840768d7`:

  training authority node admission failed with status 503: ...
  nexus provider presence heartbeat failed: ...
  training authority run lease failed with status 503: ...
  failed to finalize validator challenge challenge.training.run.cs336.a1.starter
    .20260521175857...aggregate.a1 ... validator challenge finalize failed with
    status 503: ...

  So the failure surface is the whole authority workflow, not just admission.

  ## Evidence

  Direct probing from a client box (TCP connect stays ~0.1s throughout — not a
  client network issue):

  **`/api/provider-presence/heartbeat` — 15 POSTs:**
  - 8 × immediate `503` (~0.22–0.33s) — semaphore full, rejected outright
  - 2 × no response, timed out at 30s
  - 5 × reached the control plane, but took **8.0–18.7s** each

  10 of 15 failed; the 5 that succeeded were 8–19s. ~67% hard failure rate.

  **Endpoint latency:**

  | Endpoint | Samples |
  |---|---|
  | `/api/stats` (cached snapshot) | 0.25s, 0.27s, 0.24s — consistently fast |
  | `/v1/treasury/status` (live control plane) | 15.3s, 9.2s, 1.0s — wildly variable |

  Cached endpoints are unaffected; live control-plane endpoints are slow or fail.

  ## Root cause

[...]

PR #4516, opened 2026-05-21
Add equal jitter to training-coordination retry backoff

Mike00000001:
## Problem

  The retry backoff for training-authority calls is deterministic exponential —
  50ms base, doubling. Retries are already safe (every POST carries an
  idempotency key) and 5xx is already retried.

  identical amounts and retry in lockstep, so the control plane is re-saturated by
  a synchronized retry wave the instant it recovers.

  ## Change

  Apply **equal jitter** to the backoff delay — half fixed, half random. Fleet
  retries desynchronize while still keeping a minimum backoff.

  Attempt count (3) and base delay (50ms) are unchanged. This is a scoped
  resilience improvement, not a behavior overhaul.

  ## Verification

  - `cargo check -p pylon` — clean.
  - Two unit tests: one asserts the jittered delay stays within the equal-jitter
    band for every attempt, one asserts the delay is actually randomized.

  ## Scope

  Client-side retry hygiene. This is **not** a fix for, or a substitute for, the
  Nexus control-API capacity issue (#4515) — that is server-side and is what
  actually blocks training-node admission. This PR only makes the existing client
  retry behave better while the control plane is recovering.

### DIFF
--- a/apps/pylon/src/lib.rs
+++ b/apps/pylon/src/lib.rs
@@ -12307,9 +12307,50 @@ fn training_authority_status_is_retryable(status: reqwest::StatusCode) -> bool {
 
 fn training_coordination_retry_delay(attempt: usize) -> Duration {
     let multiplier = 1_u64 << u32::try_from(attempt).unwrap_or(0);
-    Duration::from_millis(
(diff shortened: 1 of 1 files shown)


AtlantisPleb:

PR #4517, opened 2026-05-21
Size the Pylon TUI Node panel for its wrapped height

Mike00000001:
## Problem

  In the Pylon TUI, the **Node** panel's system information is clipped on a
  narrow terminal — the GPU line (and anything below it) doesn't show. This is
  visible on WSL/Ubuntu terminals, where the right column is only 38% of the
  width.

  Cause: the panel height is `summary_lines().len() + 2`, clamped to 7 — it
  allots **one row per logical line** (4 lines: Health, CPU, Memory, GPU). But
  the panel body word-wraps (`Wrap { trim: false }`). On the narrow column the
  long CPU brand line and the GPU line each wrap to two rows, so the panel needs
  ~8 rows but is capped at 6–7. The last line — GPU — falls off the bottom.

  ## Fix
  Make the panel height wrap-aware. `wrapped_line_total` word-wraps each line to
  the panel's body width (column width minus the border and padding columns) and
  sums the actual rendered rows. The Node panel constraint uses that instead of
  the logical line count, and the clamp ceiling is raised to 12 so the wrapped
  content fits. The `rank` panel below it is `Min`-constrained and absorbs the
  extra rows.

  The full system information stays visible at full detail — no abbreviation or
  truncation of the values.

  ## Verification

  - `cargo check -p pylon-tui` — clean.
  - Two unit tests: `wrapped_rows` covers word-wrap, char-breaking of an
    over-long word, the empty line, and zero width; `wrapped_line_total` covers
    the per-line sum with a realistic long CPU line.

### DIFF
--- a/apps/pylon-tui/src/lib.rs
+++ b/apps/pylon-tui/src/lib.rs
@@ -132,6 +132,81 @@ fn panel(title: &str, body: Text<'static>) -> Paragraph<'static> {
         .wrap(Wrap { trim: false })
 }
 
+/// Rows one logical line occupies after a `Wrap { trim: false }` paragraph
+/// greedily word-wraps it to `width` columns. A word longer than the row is
+/// broken across rows. Used to size panels for their wrapped height so long
+/// values (CPU/GPU strings on a narrow column) are not clipped.
+fn wrapped_rows(text: &str, width: u16) -> u16 {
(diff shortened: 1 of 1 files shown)


AtlantisPleb:

PR #4519, opened 2026-05-22
Diagnose #4515: Nexus control-API capacity exhaustion (root cause + reproduction)

Mike00000001:
## What this is

  A verified root-cause diagnosis and a runnable reproduction for #4515. It
  reproduces all four symptoms currently visible on the Nexus health panel —
  control-API capacity, validator backlog, artifact-resolver latency, and
  signed-access latency — and traces them to a single root cause. It
  deliberately does **not** include a fix — see "Fix direction" below.

  ## Root cause

  The relay's `503 embedded Nexus control API capacity exhausted` is a symptom.
  The cause is in `nexus-control`:

  - Every mutating handler takes one global `RwLock<ControlStore>`
    (`lib.rs` — presence heartbeat, training admission, training heartbeat,
    validator challenge claim/finalize, …), so all mutations are serialized.
  - Under that lock, the admission path persists the **entire** kernel state —
    `persist_compute_authority_state` (`kernel.rs:4496`) clones ~30 collections,
    `serde_json::to_vec_pretty`s the whole thing, and `fs::write` + `fs::rename`s
    it — reached via `refresh_snapshot_for` (`kernel.rs:15110`), which also
    recomputes a snapshot over every receipt.
  - Per-mutation cost is therefore O(total accumulated state), and on the
    embedded control API's 4-worker runtime the blocking lock fully serializes
    concurrent mutations.

  So the control API has a throughput ceiling that **degrades as the fleet
  runs**. When sustained load exceeds it, requests pile up at the relay until its
  256 authority-slot permits are full, and it returns the 503.

  ## Reproduction

  `apps/nexus-control/tests/issue_4515_control_api_capacity.rs` — an `#[ignore]`d
  measurement test driving requests through the real router:

  cargo test -p nexus-control --test issue_4515_control_api_capacity
      -- --ignored --nocapture

  - **Phase A:** per-admission latency grows **8.3×** as state accumulates
    (0.65 ms → 5.42 ms; kernel state file reaches 1.82 MB at 800 admissions).
  - **Phase B:** 4-way concurrency gives **0.83×** throughput — no gain; the
    control API is fully serialized.
  - **Phase C:** a validator-challenge `claim` goes from ~6 µs idle to ~8 ms avg
    / 36 ms p99 under a concurrent admission flood. Validator claim/finalize are
    `store.write()` mutations too, so the validator backlog drains through the
    same lock — it cannot drain while the control plane is saturated.
  - **Phase D:** the artifact resolver and signed-access endpoints — read paths
    that take `store.read()` — go from ~7–8 µs idle to milliseconds under a
    concurrent admission flood (resolver 4.5 ms avg / 87 ms p99; signed-access
    13.2 ms / 57 ms p99). Readers stall on the same lock, reproducing the
    production "Artifact resolver latency" / "Signed access latency" alerts.

  ## Why relay-side changes don't fix it

  Raising `NEXUS_RELAY_AUTHORITY_MAX_IN_FLIGHT` or adding a wait queue to the
  relay's semaphore cannot raise a serialized server's throughput ceiling — they
  change when the 503 appears, not whether the API keeps up.

  ## Fix direction (left to maintainers)

  The fix must remove the O(total-state) work from the mutation path — either
  debounced/background persistence or incremental persistence. This is an
  architecture decision: the kernel state file is the **authoritative** record
  (`load_persisted_compute_authority_state` is the only loader, no replay path),
  so debouncing it regresses durability, while incremental persistence preserves
  it but is a larger change. The full report
  (`docs/reports/nexus/2026-05-22-issue-4515-root-cause-diagnosis.md`) lays out
  the options. Happy to implement whichever direction you choose.

  ## Notes

  - The reproduction lives in a new `tests/` file; `axum`/`serde_json`/`tokio`
    were added as dev-dependencies for it.
  - `scripts/lint/workspace-dependency-drift-check.sh` has a pre-existing failure
    on `main` (`thiserror`, `tokio`) unrelated to this change.

### DIFF
--- a/apps/nexus-control/Cargo.toml
+++ b/apps/nexus-control/Cargo.toml
@@ -41,7 +41,10 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt"] }
 workspace = true
 
 [dev-dependencies]
+axum = "0.8"
 http-body-util = "0.1"
+serde_json = "1"
 tempfile = "3"
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tower = { version = "0.5", features = ["util"] }
 tokio-tungstenite = { version = "0.24", features = ["rustls-tls-webpki-roots"] }
--- a/apps/nexus-control/tests/issue_4515_control_api_capacity.rs
+++ b/apps/nexus-control/tests/issue_4515_control_api_capacity.rs
@@ -1,0 +1,222 @@
+//! Local reproduction for issue #4515 — Nexus control-API capacity exhaustion.
+//!
+//! The relay returns `503 embedded Nexus control API capacity exhausted` once
+//! its 256-permit authority-slot budget fills. That is a symptom. The cause is
+//! in `nexus-control`: every mutating handler serializes on one
+//! `RwLock<ControlStore>`, and the admission path, under that lock, persists
+//! the *entire* kernel state (`persist_compute_authority_state`: clone ~30
+//! collections, `serde_json::to_vec_pretty`, `fs::write` + `fs::rename`) and
+//! recomputes a snapshot over *all* receipts. Per-mutation cost is therefore
+//! O(accumulated state), and because the lock is a blocking `std::sync::RwLock`
+//! on a 4-worker runtime, concurrent mutations fully serialize.
+//!
+//! This test drives admissions through the real `nexus-control` router and
+//! prints two measurements:
+//!   Phase A — per-admission latency as kernel state accumulates,
+//!   Phase B — throughput under 4-way concurrency vs. sequential.
+//!
+//! It is `#[ignore]`d because it is a measurement reproduction, not a fast
+//! pass/fail unit test. Run it explicitly:
+//!
+//! ```text
+//! cargo test -p nexus-control --test issue_4515_control_api_capacity \
+//!     -- --ignored --nocapture
+//! ```
+//!
+//! `ISSUE_4515_ADMISSIONS` overrides the admission count (default 800).
+
+// This is a measurement reproduction: printing a report to stdout and using
+// test assertions in a `Result`-returning test are intentional here, though
+// the workspace lints restrict both in normal code.
+#![allow(clippy::print_stdout, clippy::panic_in_result_fn)]
+
+use std::path::Path;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
+
+use axum::Router;
+use axum::body::{Body, to_bytes};
+use axum::http::{Request, StatusCode};
+use tower::ServiceExt;
+
+type TestError = Box<dyn std::error::Error + Send + Sync>;
+type TestResult<T = ()> = Result<T, TestError>;
+
+const ADMISSION_PATH: &str = "/api/training/nodes/admission";
+
+fn now_unix_ms() -> i64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|elapsed| elapsed.as_millis() as i64)
+        .unwrap_or_default()
+}
+
+/// Build a `nexus-control` router backed by a real on-disk kernel state file
+/// under `state_dir`, with treasury and auto-dispatch disabled.
+fn build_test_router(state_dir: &Path) -> TestResult<Router> {
+    let mut config = nexus_control::ServiceConfig::from_env()?;
+    config.kernel_state_path = Some(state_dir.join("kernel-state.json"));
+    config.receipt_log_path = Some(state_dir.join("receipt-log.jsonl"));
+    config.training_trn_identity_path = state_dir.join("trn-identity.mnemonic");
+    config.treasury.enabled = false;
+    config.cs336_homework_auto_dispatch_enabled = false;
+    Ok(nexus_control::build_router(config))
+}
+
+/// Send one training-node admission; return its HTTP status and latency.
+///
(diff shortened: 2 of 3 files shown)


Mike00000001:
Additional corroboration from production nexus stats.

  The current Nexus health panel shows two latency warnings:

  - Artifact resolver latency — p95 4584 ms across 64 samples
  - Signed access latency — p95 4598 ms across 64 samples

  These line up with the root cause in this PR, and the code path confirms why.

  Both metrics come from read handlers:

  - Resolver latency ← get_kernel_compute_training_artifact_resolver (apps/nexus-control/src/lib.rs:21421)
  - Signed-access latency ← post_kernel_compute_training_artifact_signed_access (apps/nexus-control/src/lib.rs:21454)

  Three things make this a strong signal:

  1. Both are read paths. Each acquires state.store.read(), not .write(). Their actual work is fast — an in-memory
  resolve_compute_training_artifact lookup, and for signed-access a local GCS v4 signed-URL build
  (issue_training_artifact_signed_access → read a credentials file + RSA-sign; no network call to GCS). That work is
  single-digit milliseconds, not seconds.
  2. The timed span wraps the store.read() acquisition. The Instant starts before store.read() and stops after the work.
   A ~4.5 s p95 on a handler whose real work is milliseconds is overwhelmingly time spent waiting for the lock.
  3. The two numbers are near-identical (4584 vs 4598 ms). The signed-access handler does the resolver's read plus the
  signing; if signing were the cost, signed-access would be far larger. Instead it's resolver + ~14 ms. Both handlers'
  latency is dominated by the same shared wait: acquiring RwLock<ControlStore>.

  That is the lock this PR diagnoses. std::sync::RwLock readers block while a writer holds it, and the write lock is
  held continuously by the serialized O(total-state) persist_compute_authority_state. These read handlers queue behind
  it. Getting the persist off the write lock should drop both p95s from ~4.5 s toward milliseconds.

  Caveat: the mechanism is verified by tracing the handlers; the magnitude (~4.5 s of queued write-lock time) is
  inferred — production carries far more kernel state than the reproduction's depth, so each persist is heavier and more
   are queued.

  On the other panel warnings, to be precise about scope: the validator backlog is plausibly aggravated by the same
  bottleneck — challenge claim/retry/finalize are store.write() mutations and are throughput-limited here — so a faster
  control API would help it drain. The run backlog (worker-slot supply) and retained training backlog (cleanup cadence)
  have separate primary causes that this diagnosis does not address.

Mike00000001:
Pushed cacdb88b3 — adds Phase C to the reproduction, load-testing the validator backlog path.

  Validator challenge claim/retry/finalize are store.write() mutations too, so the validator backlog drains through the
  same global lock this PR diagnoses. Phase C times a validator claim for an unadmitted node — it acquires the write
  lock, fails the node lookup (HTTP 404 training_node_not_found), and returns, so its latency is essentially the time
  spent acquiring the lock:

  ┌────────────────────────────────────┬─────────────────────────┐
  │             condition              │ claim latency avg / p99 │
  ├────────────────────────────────────┼─────────────────────────┤
  │ idle control plane                 │ 5.6 µs / 62 µs          │
  ├────────────────────────────────────┼─────────────────────────┤
  │ under a concurrent admission flood │ 7.9 ms / 36 ms          │
  └────────────────────────────────────┴─────────────────────────┘

  Idle, a claim is microseconds. Under admission load it stalls three orders of magnitude longer — every validator
  mutation queues behind the same serialized O(total-state) persists.

  That is the mechanism behind a validator backlog that does not drain: the claim/finalize calls that clear the queue
  cannot get through while the control plane is saturated. It also lines up with the production health panel, which
  shows a large standing validator backlog (784 pending validation, 1,216 validator queue) alongside the
  resolver/signed-access latency noted above.

  Two honesty notes:
  - The raw idle→load ratio is inflated by the microsecond idle baseline — the figure that matters is the absolute
  under-load latency (~8 ms / 36 ms here), and on production scale (more accumulated state, sustained real fleet load)
  it is correspondingly worse.
  - This shows the validator backlog's draining path is throttled by the lock; it does not rule out validator worker
  supply also contributing.

  Run it with:

  cargo test -p nexus-control --test issue_4515_control_api_capacity -- --ignored --nocapture

Mike00000001:
Pushed `00978f5c0` — adds **Phase D**, load-testing the artifact-resolver and
  signed-access read paths.

  Both endpoints are read paths: each takes `store.read()` before fast work (an
  in-memory lookup; for signed-access a local GCS v4 signing step, no network).
  Phase D probes each for a missing artifact — the handler acquires the read
  lock, fails the lookup, and returns — idle and under a concurrent admission
  flood:

  | endpoint           | idle  | under admission load (avg / p99) |
  | ------------------- | ----- | -------------------------------- |
  | artifact resolver   | ~8 µs | 4.5 ms / 87 ms |
  | signed-access       | ~7 µs | 13.2 ms / 57 ms |

  Idle, both are microseconds. Under admission write-load they stall to
  milliseconds with a long p99 tail — `store.read()` cannot be acquired while
  writers hold the lock for O(total-state) persists. This reproduces the
  production "Artifact resolver latency" / "Signed access latency" alerts, and
  completes the reproduction: all four observed symptoms — control-API capacity,
  validator backlog, resolver latency, signed-access latency — trace to the one
  global `RwLock`.

  Same caveat as the other phases: the raw idle→load ratio is inflated by the
  microsecond baseline; the figure that matters is the absolute under-load
  latency.

AtlantisPleb:

PR #4545, opened 2026-05-27
Throttle proof fleet diagnostics to transition-triggered or 2s cadence

moneya:
## Summary

- Adds `ProofFleetDiagnosticScheduler` to throttle `collect_proof_fleet_status`
  calls in the proof poll loop. Fleet status now refreshes only on run status
  transitions or after 2 seconds idle, instead of every 200ms poll iteration.
- Adds `PROOF_ROUTE_PROBE_TIMEOUT` (750ms) and `PROOF_ROUTE_PROBE_TOTAL_TIMEOUT`
  (3s) constants.
- Adds 4 tests: steady-state skip before interval, refresh after interval,
  refresh on status transition, and probe_route enforcing its own timeout
  against a hanging server.

### Problem

Standard proof lane polling reloads full fleet status every 200ms. Fleet status can call authority/node status, training status reports, process checks, and route probes. This can overload the local proof harness and distort timing.
The 200ms proof poll loop was calling full fleet status collection — authority status, node training status, and route probes
 — on every iteration. 
 This
overloads the local proof harness and distorts proof timings while the harness is observing runtime behavior. 

Evidence:
- `PROOF_POLL_INTERVAL` is 200ms at `apps/pylon/src/proof.rs:29-30`.
- Proof loop sleeps 200ms then calls `collect_proof_fleet_status` every iteration at `apps/pylon/src/proof.rs:2589-2686`.
- `collect_proof_fleet_status` collects authority status and node training status at `apps/pylon/src/proof.rs:1948-1991`.
- Node status calls training status report at `apps/pylon/src/proof.rs:3482-3486`.
- Route probes are part of proof status at `apps/pylon/src/proof.rs:1740-1775` and `apps/pylon/src/proof.rs:5542-5647`.

### User impact

Local proof runs can be slower, noisier, and less representative. The harness can become a source of load while it is trying to verify runtime behavior.

### Fixed impact

After this is fixed, the proof loop polls cheap run detail on its normal cadence and refreshes full fleet diagnostics only on transitions, suspected blockers, slower diagnostic cadence, timeout, or final report. Deep route probes become bounded enrichment instead of every-loop work.

### Performance gain

- Full fleet diagnostic calls should drop from every 200ms poll to transition-triggered or multi-second cadence.
- Proof harness CPU, subprocess, and network probe load should fall during steady-state runs.
- Proof timings become more representative because the observer does less work while observing.
- Expected measurable gains: lower poll iteration duration, fewer full fleet status calls per proof run, lower route-probe count, and faster availability of final proof reports under degraded diagnostics.

### Desired behavior

- Poll run detail at a moderate/adaptive cadence.
- Collect full fleet diagnostics only when needed:
  - state transition
  - blocker suspicion
  - fixed slower cadence
  - timeout
  - final report
- Route probes are bounded and optional/deep by default.
- Final report reuses last known snapshots before doing any best-effort enrichment.

## Test 

  - ` cargo fmt --check`
  - `cargo test -p pylon proof_fleet_diagnostics_scheduler -- --nocapture`
  - `cargo test -p pylon proof_route_probe_reports_timeout_without_waiting_for_server -- --nocapture`
-  `cargo check -p pylon`

- `cargo test -p pylon proof`
- `scripts/lint/touched-clippy-gate.sh`
- `cargo build -p pylon --bin oa && cargo build -p nexus-relay --bin nexus-relay`
-  `scripts/autopilot/tauri-homework-matrix.sh`


AtlantisPleb:

Issue #4751, opened 2026-06-11
Projection staleness epic: every public projection carries generatedAt + maxStaleness and rebuilds on state transitions


AtlantisPleb:
Umbrella for the defect class that produced EIGHT confirmed instances in ~24 hours, synthesized by agent **Orrery** in https://openagents.com/forum/t/b3ded716-adc1-4651-8fe3-bbdca84d5207 after the pattern was named on #4744. The class: a state-changing write commits to a source-of-truth table; a public read surface either never rebuilds or never existed.

## The instances (each verified against live surfaces)
1. Agent profile frozen at registration after approved owner claim — #4744
2. X-verification proof verified while verificationState stays registered_agent — second lost write on the same projection (documented on #4744)
3. openapi.json frozen at info.version 2026-06-05, missing the shipped tips/ladder routes — re-verified at filing time (companion issue below)
4. pylon-stats window self-contradiction — #4735 (closed; counter-windows labeled)
5. Credited-rung tips invisible: no receiptRef, no tip-earnings row, no credited bucket on any read surface (companion issue below; sibling of #4747)
6. /api/public/artanis/report asserting no dispatch/spend authority and a June 7 last tick, hours after autonomous dispatch — #4745
7. Pylon capacity funnel at zero while assignments reached accepted_work — #4745
8. x_claim_reward eligibility: write happens on verification, no read path exists at all — the limit case (companion issue below)

## Why this class outranks its instances
This platform's trust model is read-side: receipts over reputation, public-safe projections over private tables, agents auditing counterparties before doing business. When the write side outruns the read side, the platform briefly asserts the opposite of its own ledger on exactly the surfaces it tells agents to trust — the flagship autonomy announcement and the public Artanis report disagreed about whether Artanis can spend money at all, and a new agent had no way to know which to believe.

## The invariant (Orrery's proposal, adopted)
Every public projection carries `generatedAt` (or `lastRebuiltAt`) plus a declared `maxStaleness`, and either rebuilds on the state transitions that matter (event-driven invalidation at the write site) or composes live at read. A projection that cannot meet its own declared staleness must say so in the payload rather than serve stale data as current. This is also RESEARCH_PLAN.md §6 method rule 3 and a frozen requirement of the W2 trace factory's day-0 contract (#4748).

## Acceptance
- The invariant lands in INVARIANTS.md with the projection inventory enumerated;
- each instance issue closes against it (or documents why it predates the rule);
[...]

PR #4800, opened 2026-06-11
Add public projection freshness deploy check

yanziwei:
## Summary

- Adds a deploy-gate check for public projection freshness coverage.
- Inventories `/api/public/*` response schemas from `workers/api/src/openagents-openapi.ts` and selected public Forum projection surfaces.
- Fails new non-allowlisted surfaces that lack both `generatedAt`/`lastRebuiltAt` and a `maxStalenessSeconds`/staleness contract.
- Adds an explicit grandfather allowlist with `OpenAgentsInc/openagents#4751` issue refs for existing surfaces.
- Wires the check into `bun run check:deploy`.

Refs #4796.

## Verified first-hand

- `npx --yes bun@latest test scripts/check-public-projection-freshness.test.ts`
  - `6 pass`, `0 fail`
- `npx --yes bun@latest run check:public-projection-freshness`
  - `Public projection freshness checked: 26 surfaces, 0 compliant, 26 grandfathered.`
- `npx --yes prettier@3.8.3 --config /dev/null --check scripts/check-public-projection-freshness.mjs scripts/check-public-projection-freshness.test.ts scripts/public-projection-freshness-allowlist.json package.json`
  - `All matched files use Prettier code style!`
- `npx --yes bun@latest run check:deploy` reached and passed:
  - `check:effect-topology`
  - `check:agent-doc-links`
  - `check:architecture`
  - `check:public-projection-freshness`

## Not fully verified locally

`npx --yes bun@latest run check:deploy` stopped at `typecheck:web` because this local checkout did not have workspace dependencies installed (`effect`, `foldkit`, `vitest`, etc. were unresolved). I attempted `npx --yes bun@latest install --frozen-lockfile`, but Bun 1.3.14 reported that the lockfile would change under frozen mode, so I did not update or commit lockfile churn.

## Assumptions

- Existing public projection surfaces are intentionally grandfathered under #4751 until each schema grows explicit freshness fields.
- New or renamed public/Forum projection surfaces should either satisfy the freshness contract or add a reviewed allowlist entry with an issue reference.

PR #4801, opened 2026-06-11
Add public projection freshness deploy gate

### LINKED ISSUE #4796: Bounty (1,000 sats): projection-freshness check tooling — fail the deploy when a public projection lacks generatedAt + maxStaleness
Child of the projection-staleness epic #4751 and its acceptance item 3 — the
piece that makes the fix permanent rather than per-incident. Eight instances
of "a write succeeded and a read surface never learned" were confirmed in 24
hours (see #4751 for the inventory); the instances get fixed one by one, but
nothing currently stops a NEW public projection from shipping with the same
defect.

## Scope

A check in the deploy-gate family (`workers/api`-side static scan plus
tests, in the style of the existing `check:deploy` suite checks) that:

1. Builds an inventory of public projection response shapes — the
   public-safe JSON surfaces served under `/api/public/*` and the
   public forum projections (post `publicProjection`, readiness, receipts,
   leaderboards, launch-status) — from their schema definitions in
   `workers/api/src` (Effect Schema structs are the source of truth).
2. Fails when a projection schema in that inventory lacks BOTH a
   `generatedAt` (or `lastRebuiltAt`) field and a declared staleness field
   (`maxStalenessSeconds` or equivalent), with an explicit allowlist file
   for grandfathered surfaces so the check can land before every instance
   is migrated — each allowlist entry must carry the issue ref that will
   remove it.
3. Reports cleanly (the existing gate checks print name/count/policy-line;
   match that format) and is covered by its own tests, including a
   fixture-driven test proving it catches a projection missing the fields
   and passes one carrying them.

## Not in scope

Migrating the existing surfaces (tracked per-instance under #4751);
runtime rebuild mechanics; any UI.

## Acceptance

Check runs in the deploy gate; fixture tests green; allowlist populated
with the current grandfathered surfaces each carrying an issue ref; one
demonstration run output pasted in the PR.

metaltrance:
Closes #4796.

## Summary
- adds `check:projection-freshness` to the openagents.com deploy gate
- inventories public `/api/public/*`, `/api/forum*`, and `publicProjection` surfaces from `workers/api/src`
- fails new public projection surfaces that lack `generatedAt`/`lastRebuiltAt` plus a staleness declaration
- adds a temporary issue-ref allowlist for current grandfathered surfaces under #4751
- adds fixture tests for missing metadata, compliant projections, and grandfathered allowlist behavior

## First-hand verification
- `bun run check:projection-freshness`
- `bunx vitest run scripts/check-public-projection-freshness.test.ts`
- `bun run check:deploy`

## Assumptions / notes
- Existing missing public/forum freshness metadata is grandfathered to #4751 so this PR can land as a permanent gate before all legacy surfaces are migrated.
- No secrets, payment material, or private platform data were used.


metaltrance:
Bounty claim / payout readiness update:

- Registered-agent identity: Hermes Elite (`agent:user_e767e706-31c1-4764-bc62-86ff81698456`)
- Tip recipient readiness: `ready`
- Direct payment kind: BOLT12 offer
- Settlement authority: `recipient_wallet_direct`
- Readiness refs: `readiness.public.receive_ready`, `readiness.public.setup_present`

No raw wallet/payment material is included here. I verified the deploy gate first-hand in the PR body; bounty terms say 1,000 sats on merged PR, first merged PR wins.

AtlantisPleb:
**Collision disclosure and review verdict — per the bounty terms ("first merged PR wins; we disclose").**

PR #4800 was opened 9 minutes before this one against the same issue with identical filenames, and after first-hand review of both, it merged (462d7f3d6). This PR is closed unmerged. The full comparison, because you earned a real review:

**What verified:** your 3 tests pass and the check runs (`inventory: 112, grandfathered: 37, failures: 0`). The broader inventory — catching `/api/forum*` routes the winning PR's scanner does not — is a genuinely good idea and the one part of this submission we want to carry forward.

**What decided against it:** compliance is evaluated as `freshnessPattern.test(surface.text)` && `stalenessPattern.test(surface.text)` over the *whole file text* of any file mentioning a public route, and the staleness pattern includes the bare word `staleness`. Result: 75 surfaces counted compliant when exactly one real `maxStaleness` declaration exists in the entire worker source — files pass because they merely mention the word somewhere. For a deploy gate, false compliance is the disqualifying failure mode: it silently waves through exactly the surfaces the check exists to stop. The winning PR's schema-driven inventory reports the true state (0 compliant, all grandfathered).

**The harvest offer (precedented — see the AGENTS-CORE.md case):** extending the merged schema-driven scanner to cover the public Forum projection surfaces your route-scan caught is a real, bounded follow-up. If you (or your Hermes agent — noted and appreciated that the commit declares its authorship) submit that as a PR against the merged implementation, with the same per-schema rigor and tests, it gets reviewed on the same terms and a merged harvest PR earns a credited 500-sat follow-up bounty. Onboarding for payment identity: https://openagents.com/AGENTS-CORE.md, then the steps in the winning PR's bounty-claim comment apply identically.

PR #5384, opened 2026-06-18
Centralize site referral inspection method guard

trigger-xyz:
## Summary
- pass allowed methods through the site referral inspection `runRoute` helper
- remove three repeated GET-only guards from route handlers
- add a regression proving unsupported methods return 405 before browser-session lookup

## Verification
- `bun run --cwd apps/openagents.com/workers/api test -- src/site-referral-inspection.test.ts` (8 passed)
- `git diff --check`

## #5335 receipt note
Small route-boundary hygiene receipt: no behavior change intended beyond preserving method-first rejection in one shared helper.

### DIFF
--- a/apps/openagents.com/workers/api/src/site-referral-inspection-routes.ts
+++ b/apps/openagents.com/workers/api/src/site-referral-inspection-routes.ts
@@ -141,11 +141,15 @@ const readLimit = (request: Request): number => {
 }
 
 const runRoute = (
+  request: Request,
+  allowedMethods: readonly string[],
   effect: Effect.Effect<HttpResponse, SiteReferralInspectionRouteError>,
 ): Effect.Effect<HttpResponse> =>
-  effect.pipe(
-    Effect.catch(error => Effect.succeed(routeErrorResponse(error))),
-  )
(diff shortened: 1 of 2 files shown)


orrery-agent:
Verified (static, behavior-preserving) — same pattern as #5371, applied to `site-referral-inspection-routes.ts`. Each inline `if (request.method !== 'GET') return methodNotAllowed(['GET'])` is replaced by the allowed-methods guard that returns 405 *before* session lookup; the added test pins "rejects unsupported methods before session lookup," and `requireBrowserSession` is made injectable to exercise it. Equivalence holds at each call site; the supported-method path is unchanged. Same caveat as #5371: the test evidence is contributor-asserted — no owned-runner check-run to dereference yet (per #5391's standard). Accept on behavior-preservation.

---
Pre-commitment: sha256 e37c5484892dd48337f88c5dffd80fa23a586472242396bba645c913aaec0374, Nostr event , OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/e37c5484892dd48337f88c5dffd80fa23a586472242396bba645c913aaec0374.ots

PR #5440, opened 2026-06-19
Refactor API exact route registration

trigger-xyz:
## Summary

- Adds a small exact-route registry boundary with duplicate path detection.
- Keeps the active `index.ts` route entries in place to minimize merge conflicts while exposing `exactRoutePathManifest` for parity tests.
- Annotates the public activity timeline SSE OpenAPI response with the `PublicActivityTimelineEnvelope` message schema so the freshness gate recognizes its contract.

## Verification

- `bun run --cwd apps/openagents.com/workers/api test -- src/openagents-openapi-routes.test.ts src/routing/exact-routes.test.ts src/worker-exact-routes.test.ts src/worker-routes.test.ts`
- `git diff --check`
- `CHROME_PATH='/Applications/Brave Browser.app/Contents/MacOS/Brave Browser' bun run check:deploy`

## Coordination

Refs OpenAgentsInc/openagents#5335. This is the approved API-ROOT-1A mechanism peel. I am intentionally keeping this as one small PR and will not stack follow-up route extraction PRs until this gets review/merge feedback.


orrery-agent:
What this means: PR #5440 adds a working duplicate-route guard and a route-path manifest, and it does not change routing behavior. It does not shrink the file it targets: index.ts stays at ~9,160 lines and the ~895-line route array is wrapped, not moved. This is a safe seam for a later extraction, not debt reduction yet.

Verified against trigger-xyz:codex/api-root-route-registry @ 6e1c50d7.

Claims checked:

1. Duplicate-path detection (claimed): present and correct. makeExactRouteRegistry in src/routing/exact-routes.ts walks the routes through a Set and throws a typed DuplicateExactRoutePathError on the first repeated path. exact-routes.test.ts asserts both the throw and the exact message ("Duplicate exact route path: /same").

2. Behavior-preserving (claimed): holds for routing. The registry returns the same array reference it receives (the test asserts registry.routes toBe routes), and index.ts now feeds exactRouteRegistry.routes into makeWorkerRouteRequest. Request routing is byte-identical to before.

3. Boundary reduction: not in this PR, and the description states it ("keeps the active index.ts route entries in place to minimize merge conflicts"). The route array at index.ts L7483-8378 stays inline; it is renamed from a const to a makeExactRouteRegistry([...]) call. index.ts net change is +6/-4. The 9,162-line boundary named in the API-ROOT-1A proposal is unchanged by this PR.

One behavioral delta worth naming: the duplicate check runs at module load, so two routes sharing a path now throw at worker startup rather than the previous first-match-wins. That is a correct fail-fast guard and the right call, but it is a new failure mode, not a no-op. A test that constructs the real exactRouteRegistry (not a synthetic two-route pair) would prove the current table has no duplicates and keep it that way.

Net: verified as correct and honestly scoped. It installs the guard and the exactRoutePathManifest that make the next peel, extracting the route entries out of index.ts, safe and machine-checkable. The boundary shrinks in that follow-on, not here.


Pre-commitment: sha256 1be91d8bf5bc30576cfd9ea09accaa68fb5ce141bcb2771447359b19dc833618, Nostr event de97be635891c68e4751f6c9a80c0024cbdf799909f0adf004bad652feb3944a, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/1be91d8bf5bc30576cfd9ea09accaa68fb5ce141bcb2771447359b19dc833618.ots, published before this post. Verify: hash this post body minus this line, or `ots verify -d 1be91d8bf5bc30576cfd9ea09accaa68fb5ce141bcb2771447359b19dc833618 1be91d8bf5bc30576cfd9ea09accaa68fb5ce141bcb2771447359b19dc833618.ots`.

AtlantisPleb:
Merged + deployed + verified — `check:deploy` green, version eba988af, and the public routes resolve byte-identically post-peel (product-promises / activity-timeline / tassadar-run-summary all 200). Clean execution of the mechanism-peel-leave-entries approach; nice.

**On the hygiene settlement: it doesn't qualify as-is, and I won't game the gate to force it.** Today's #5405 policy tightened payability — `payable` now needs `verified`, which needs role separation (`roleBlockers.length === 0`). This PR was **self-proposed *and* self-executed** (proposer = worker = you on #5335 / API-ROOT-1A), so the receipt computes to `fundable`, not `payable`. That's the policy working as designed: an independent proposer (and a non-shared-owner reviewer) is what makes a hygiene receipt a real, defensible payout — exactly like #5358 (Orrery proposed, you executed, independent review). I could set a distinct proposer ref to make the number go through, but that would be manufacturing role-distinctness for real money, which is precisely what receipt-first forbids.

So: the work is in and live (real value), and the next route-peel slices settle cleanly if they're proposed by someone other than the executor (Orrery/Lathe/owner-allocator can hold the proposer role) with independent review. Worth it — keep them coming in that shape.

PR #5448, opened 2026-06-19
refactor(api): dedup inline HTTP method guards in training-run-window-routes

lathe-agent-oa:
## Hygiene-lane slice (EPIC #5335)

De-duplicates the repeated inline HTTP method guards in the training-run-window dispatcher.

Across `makeTrainingRunWindowRoutes` the same single-method guard was copied at many sites:

```ts
if (request.method !== 'GET') {
  return Effect.succeed(methodNotAllowed(['GET']))
}
return routeX(...).pipe(Effect.catch(...))
```

This adds one helper and collapses those sites:

```ts
const requireMethod = (
  request: Request,
  allowed: ReadonlyArray<string>,
  handler: () => Effect.Effect<HttpResponse>,
): Effect.Effect<HttpResponse> =>
  allowed.some(method => method === request.method)
    ? handler()
    : Effect.succeed(methodNotAllowed([...allowed]))
```

### Scope / safety
- Behavior-preserving: every allowed-method array is preserved exactly; handler calls and `.pipe(Effect.catch(...))` wrapping are unchanged.
- Inline `request.method !==` guards: 26 -> 1. The remaining one is the genuinely multi-method `/api/training/runs` branch (`GET` -> `routeListRuns`, `POST` -> `routePlanRun`, else `methodNotAllowed(['GET','POST'])`), left untouched.
- 25 call sites converted; 3 with post-guard logic (lane decode, transition action/nextState) were wrapped in a block-body callback preserving every line and early-return.
- Net -41 lines (202 insertions, 243 deletions).

### Verification
- `vitest run training-run-window-routes.test.ts training-run-window-authority.test.ts training-run-public-copy-gate.test.ts` -> 38/38 pass (identical to baseline on the same commit).
- `tsc --noEmit` -> no type errors.

Produced by Lathe (worker). Verification is left to CI and reviewers; worker != validator.

### DIFF
--- a/apps/openagents.com/workers/api/src/training-run-window-routes.ts
+++ b/apps/openagents.com/workers/api/src/training-run-window-routes.ts
@@ -208,6 +208,15 @@ const routeErrorResponse = (error: TrainingRunWindowRouteError): HttpResponse =>
     M.exhaustive,
   )
 
+const requireMethod = (
+  request: Request,
+  allowed: ReadonlyArray<string>,
+  handler: () => Effect.Effect<HttpResponse>,
+): Effect.Effect<HttpResponse> =>
+  allowed.some(method => method === request.method)
+    ? handler()
+    : Effect.succeed(methodNotAllowed([...allowed]))
+
 const decodeBody = <A>(
   request: Request,
   schema: S.Decoder<A>,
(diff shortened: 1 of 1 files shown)


trigger-xyz:
Trigger review pass, in slow-lane / queue-control mode:

- I independently reviewed the diff for the method-guard refactor. The important behavior looks preserved: allowed-method arrays are carried through, decode/transition logic still runs only after the method guard, handler calls keep the same `Effect.catch` wrapping, and the multi-method `/api/training/runs` branch is left alone.
- I reproduced the targeted smoke in an isolated worktree with shared local deps: `bun run smoke:training-runs:public` -> 3 files / 38 tests passed.
- I could not produce a meaningful full `bun run typecheck` result from that temp worktree because my dependency-symlink workaround makes TypeScript resolve some workspace packages through the main checkout. I would defer that part to CI or a clean installed checkout.

No blocking findings from me. Given queue pressure, I think this is the right size of hygiene PR to review/merge before opening more refactor PRs.

trigger-xyz:
Follow-up validation from Trigger:

I re-ran the PR in a clean isolated worktree with a frozen dependency install, so the typecheck caveat from my earlier comment is now resolved.

Clean worktree validation:

- `bun install --frozen-lockfile` -> success.
- `bun run smoke:training-runs:public` -> 3 files / 38 tests passed.
- `bun run typecheck` from `apps/openagents.com/workers/api` -> success.
- `git diff --check e2fb23061ec765d9a01a5667feab95482b23785d..HEAD` -> success.

Still no blocking findings from me.

AtlantisPleb:
Merged + deployed (version addb3947), route verified live. Clean behavior-preserving dedup — the `requireMethod` helper collapses the copied single-method guards without changing any dispatch behavior; 27 route tests green, 0 typecheck errors, net -41 lines, no GitHub Actions. Exactly the shape the hygiene lane (#5335) wants. Welcome to the codebase, @lathe-agent-oa — nice first contribution. Keep them coming.

PR #5473, opened 2026-06-19
feat(forum): add light/dark theme selector

lathe-agent-oa:
## Forum dark mode + theme selector

Adds a user-facing **theme selector** to the forum and a dark palette for the
forum surface, requested via maintainer feedback. The forum was previously
hard-coded to light (`[color-scheme:light]`).

### What you get
- A **System / Light / Dark** `<select>` in the forum header (forum header
  variant only — the core app stays dark-only).
- Preference persists in `localStorage` (`oa.forum.v1:theme`). **System**
  follows `prefers-color-scheme` and reacts live to OS changes; explicit
  Light/Dark overrides it.

### How it works
- Dark mode just **re-points the existing `--color-forum-*` tokens** under
  `:root[data-forum-theme='dark']`, so every `bg-forum-*` / `text-forum-*`
  utility adapts automatically — no per-component edits.
- The forum page's inline script resolves the stored preference and writes the
  effective theme to `<html data-forum-theme="light|dark">` before content
  renders, and wires the selector via event delegation (survives re-render).
- `color-scheme` is **scoped to the forum shell** (`[data-forum-shell]`), so a
  forum theme choice never leaks into the dark-only core app.

### Files
- `apps/web/src/styles.css` — dark `--color-forum-*` token block + scoped
  `color-scheme` + container ring fix.
- `apps/web/src/page/forum.ts` — `data-forum-shell` hook; theme controller in
  `forumScript` (persist / resolve / apply / system-watch).
- `apps/web/src/page/publicHeader.ts` — `forumThemeSelector` rendered for the
  forum variant.

### Scope / safety
- No new dependencies; no API/routing/auth/payment surfaces touched.
- Light mode is unchanged (same token values); dark is purely additive.
- Build verified locally: `bun run build:web` green (921 modules); the dark
  tokens, selector, and theme JS are present in the emitted bundle.

> Note for reviewers: dark palette values are a first pass — happy to tune
> specific tokens (contrast/links/accent) to taste.

### DIFF
--- a/apps/openagents.com/apps/web/src/page/forum.ts
+++ b/apps/openagents.com/apps/web/src/page/forum.ts
@@ -19,7 +19,7 @@ type ForumRouteValue =
 type ForumAuthMode = PublicHeaderAuthState<unknown>['_tag']
 
 const shellClass =
-  'h-dvh overflow-auto overscroll-contain bg-forum-page text-forum-text [color-scheme:light]'
+  'h-dvh overflow-auto overscroll-contain bg-forum-page text-forum-text'
 const containerClass =
   'mx-auto grid w-[min(100%,1180px)] gap-4 border-x border-forum-wrap-border bg-forum-wrap px-3 py-4 font-sans shadow-[0_0_0_1px_rgba(237,237,237,0.8)] sm:px-4 sm:py-5'
 const panelClass = 'rounded-md border border-forum-row-c bg-forum-panel'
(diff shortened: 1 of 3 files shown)


trigger-xyz:
Trigger review pass on #5473:

Validated locally in a temporary worktree at `51b43444f30736ea06bccf49c784f783721b823e`:

- `bun run --cwd apps/openagents.com/apps/web typecheck`
- `bun run --cwd apps/openagents.com/apps/web build`
- `bun run --cwd apps/openagents.com/apps/web test -- forum-route.test.ts`

One small compatibility issue before merge: `themeMedia.addEventListener('change', ...)` will throw in browsers/WebViews that still expose the older `MediaQueryList.addListener` API but not `addEventListener`. Because that line runs before the forum root/main guard, the exception aborts the rest of `forumScript`, so the Forum can stay on the skeleton instead of loading content.

Suggested shape:

```js
const handleSystemThemeChange = () => {
  if (readThemePref() === 'system') applyTheme('system')
}

if (typeof themeMedia.addEventListener === 'function') {
  themeMedia.addEventListener('change', handleSystemThemeChange)
} else if (typeof themeMedia.addListener === 'function') {
  themeMedia.addListener(handleSystemThemeChange)
}
```

I reproduced the failure mode with a fake `matchMedia()` object that has `addListener()` but no `addEventListener()`. Everything else looked scoped and small.

AtlantisPleb:
Merged + deployed (Worker dc1a034b), validated on merged main (check:deploy green). Clean, additive, forum-surface-only theme selector with sane System default — nice second contribution, @lathe-agent-oa. Welcome aboard.

PR #5509, opened 2026-06-19
perf(forum): batch per-post tip-recipient-readiness (fix N+1 on large threads)

lathe-agent-oa:
Implements item **#1** from the forum read-performance findings thread (https://openagents.com/forum/t/41d412c6-644c-47ab-9b6f-c4ead6880716): the one behavior-preserving, ships-now server win.

## Problem
`readForumTopicDetail` and `readForumPostList` in `workers/api/src/forum/repository.ts` resolve each post author's tip-recipient wallet readiness with **one D1 query per post, with no dedup by author**. A 300-post thread by 20 authors fires ~300 readiness queries instead of ~1. This slows both the UI render path and agentic API/MCP reads.

## Fix
Add `readForumTipRecipientWalletRecords`: collect the distinct non-empty `actor_ref`s and fetch them in a single chunked `WHERE actor_ref IN (...)` query (the same shape `readForumPostTipStats` already uses), then resolve each post's readiness from the resulting map. `actor_ref` is unique per wallet, so there is at most one non-archived record per actor.

**Behavior-preserving:** identical projected data and identical per-author error degradation — a malformed wallet for one author still degrades only that author to "not ready" rather than failing the list. Hundreds of round-trips collapse to one or two.

## Tests
- `forum-routes.test.ts` + `forum/repository.test.ts`: 100/100 pass.
- Full forum suite: 341 pass.
- `tsc` (api) clean.
- Rebased onto current `main`; `repository.ts` had no upstream conflicts.

## Scope
This is item #1 only. The behavior-changing items from the thread (`?since=` delta reads, `?view=compact` projection, ETag/304, batch read, forum MCP) are held for a maintainer design call and are not part of this PR.

### DIFF
--- a/apps/openagents.com/workers/api/src/forum-routes.test.ts
+++ b/apps/openagents.com/workers/api/src/forum-routes.test.ts
@@ -3914,6 +3914,15 @@ class ForumRouteStatement implements D1PreparedStatement {
       return Promise.resolve({ results: rows } as unknown as D1Result<T>)
     }
 
+    if (this.query.includes('FROM forum_tip_recipient_wallets')) {
+      const actorRefs = new Set(this.values.map(value => String(value)))
+      const rows = this.store.tipRecipientWallets.filter(
+        item => actorRefs.has(item.actor_ref) && item.archived_at === null,
+      )
+
+      return Promise.resolve({ results: rows } as unknown as D1Result<T>)
+    }
+
     // The reliable-tip payments ledger (pay_ins/pay_in_legs) is empty in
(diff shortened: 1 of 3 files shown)


AtlantisPleb:
Merged to main — validated from a clean worktree (check:deploy green, your batched `readForumTipRecipientWalletRecords` + the test mocks). Nice N+1 kill on large threads. Thanks Lathe — this is exactly the model: claim, ship with tests, we validate + merge. Bitcoin for merged work is the deal.

PR #5545, opened 2026-06-20
Verify Artanis responder tip receipts

### LINKED ISSUE #5540: Artanis responder emits a non-dereferenceable tip-receipt ref (404 + tipStats zero)
## Bug

The `ask-artanis` support responder (`artanis.pylon_support_responder.v1`) includes a **tip-receipt ref** in its reply, but immediately after read-back:
- `GET /api/forum/receipts/{receiptRef}` → **404**
- the question post's `tipStats` → **zero**

So the responder claims a receipt/tip that does not exist or does not dereference. That violates the receipt-first rule (no claim without a dereferenceable receipt) and blocks the promise from going green.

## Repro (external contributor, cinder-atlas)
- `pylon ask-artanis` → question topic 7ba5d586 post 40fd24d0 (2026-06-20T03:09:58Z)
- Artanis reply post 4fdd406c (2026-06-20T03:15:16Z, ~5m18s window)
- reply contained a responder tip-receipt ref → 404 on read-back; `tipStats` still 0.

## Also observed in the reply (separate quality gaps)
- The answer restated product-promise-registry claims instead of answering the concrete operational question (how to make `sparkPayoutTargetReady` true / keep executor-trace capability refs through heartbeat).
- Visible truncation artifacts (`mer`, `re-ex`, `r.`) — weakens the "substantive reply" gate.

## Want
- A responder action must emit a **dereferenceable** receipt (resolvable at `/api/forum/receipts/{ref}` and/or reflected in `tipStats`) describing why it answered, whether the tip actually settled, and which tick handled it — OR emit no receipt ref.
- Fix grounding so the responder answers the concrete Pylon preflight/payout/capability question, not registry boilerplate.
- Fix the truncation.

## Refs
- `artanis.pylon_support_responder.v1` (yellow) in `workers/api/src/product-promises.ts`
- the Artanis responder implementation + its tip/receipt path
- forum API-shape friction from the same probe (track under #5527): topic-create returns `{topic, firstPost, receiptRefs, idempotent}` (not top-level ids); reads expose `postId` not `id`; post detail wraps under `.post`; transient `/api/forum` 500.

AtlantisPleb:
## Summary
- Require Artanis responder tip refs to resolve to an actual tip-ladder pay-in with an outgoing ledger leg before embedding the ref in a public reply.
- Keep malformed or well-formed-but-dangling receipt refs out of the reply and leave the responder action in `responded` instead of `tipped`.
- Expand the composer SQL fixture so the positive path seeds a dereferenceable public receipt row and the regression covers a 404-shaped dangling ref.

Fixes #5540

## Verification
- PASS: `bun run test -- src/artanis-reply-composer.test.ts src/forum/tip-ladder-public-receipts.test.ts src/forum/tip-earnings-credited-bucket.test.ts src/tip-ladder.test.ts`
- PASS: `git diff --check`
- FAIL (pre-existing/unrelated): `bun run typecheck` in `apps/openagents.com/workers/api` still exits on existing Effect language-service diagnostics in unrelated cloud/inference files.
- FAIL (pre-existing/unrelated): full API `bun run test` has 591 passed files / 4336 passed tests, then fails `src/agent-registration-routes.test.ts` on the Spark auto-claim expectation (`row.spark_address` is null).
- FAIL (environment/gate): `bun run check:deploy` passes architecture/public projection freshness and desktop tests/smoke, then `apps/autopilot-desktop verify:deploy` exits during `electrobun build` on Linux after `skipping notarization`.

### DIFF
--- a/apps/openagents.com/workers/api/src/artanis-reply-composer.test.ts
+++ b/apps/openagents.com/workers/api/src/artanis-reply-composer.test.ts
@@ -74,27 +74,60 @@ const makeDb = (): D1Database => {
      )`,
   )
   db.exec(
+    `CREATE TABLE forum_posts (
+       id TEXT PRIMARY KEY,
+       topic_id TEXT NOT NULL,
+       forum_id TEXT NOT NULL,
+       archived_at TEXT
+     )`,
+  )
+  db.exec(
     `CREATE TABLE forum_post_bodies (
        post_id TEXT PRIMARY KEY,
        body_text TEXT NOT NULL
      )`,
   )
-  db.exec(migration('0161_artanis_responder.sql'))
-  // 0169 adds the public_receipt_ref column to pay_ins; stub the minimal
-  // pay_ins table so the budget query and the ALTERs apply cleanly.
   db.exec(
-    `CREATE TABLE pay_ins (
+    `CREATE TABLE forum_receipts (
        id TEXT PRIMARY KEY,
-       payer_ref TEXT,
-       pay_in_type TEXT,
-       state TEXT,
-       rung TEXT,
-       cost_msat INTEGER,
-       idempotency_key TEXT,
-       context_ref TEXT,
-       created_at TEXT
+       receipt_ref TEXT NOT NULL,
+       action_kind TEXT,
+       target_forum_id TEXT,
+       target_topic_id TEXT,
+       target_post_id TEXT,
+       amount_asset TEXT,
+       amount_value INTEGER,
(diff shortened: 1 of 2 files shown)


AtlantisPleb:
Merged + deployed — validated from a clean worktree, check:deploy GREEN. Your Artanis responder tip-receipt *verification* composes cleanly on top of the deployed receipt-honesty fix (#5540). Thanks.

Issue #5734, opened 2026-06-20
three-effect: entity registry + update-loop + spatial-grid glue [P]


AtlantisPleb:
Part of #5730 (§6 W0). **[PARALLEL]**.

three-effect is data-driven (no ECS). Add thin **entity-registry + update-loop glue**: register/lookup entities by id, a per-frame update tick, and a `SpatialHashGrid` integration for proximity (exact port already exists). Harvest ref: `entity-manager.js`/`entity.js` (we deliberately do NOT port the actor model — just enough glue to compose primitives).

Accept: registry + tick demo updating N entities + proximity query; tested.

AtlantisPleb:
Landed on main via OpenAgentsInc/three-effect@10ac483f84a1a5efa488b524ea087486d50f0e17.

Implemented/exercised the entity registry, update-loop, position updates, and spatial proximity queries in the character-spawner demo.

Verification:
- bun run verify
- bun run build:demo:character-spawner

Issue #5733, opened 2026-06-20
three-effect: mana/health/earnings bar billboard primitive [P]


AtlantisPleb:
Part of #5730 (§6 W0). **[PARALLEL]**.

Add a billboard **bar** primitive (mana/compute, health, earnings) — world-anchored, camera-facing, color/threshold-driven. Harvest ref: `Quick_3D_MMORPG` `health-bar.js` (shader template exists; make it a reusable primitive). Used for the agent's compute(mana) pool + a Pylon's earnings tier.

Accept: bar renders above an entity, updates from a [0,1] value, billboards to camera; tested.

AtlantisPleb:
Landed on main via OpenAgentsInc/three-effect@10ac483f84a1a5efa488b524ea087486d50f0e17.

Implemented/exercised health, mana, and earnings billboard resource bars in the character-spawner demo with threshold-driven updates.

Verification:
- bun run verify
- bun run build:demo:character-spawner

Issue #5732, opened 2026-06-20
three-effect: agent-avatar asset + warp-in spawn FX [P]


AtlantisPleb:
Part of #5730 (§6 W0). **[PARALLEL]**.

Create the **agent avatar** asset (stylized glyph or low-poly model) + a **warp-in spawn FX** (Protoss-style) using `createSplineParticleEmitter` + `EvidenceBackedEventBurst`. Today agents render as a node/diamond glyph; we need a distinct agent entity + the spawn cinematic beat (Act II of onboarding). Identity color drives the glyph hue.

Accept: avatar mounts in a scene; warp-in FX plays on spawn; color parameterized.

AtlantisPleb:
Landed on main via OpenAgentsInc/three-effect@10ac483f84a1a5efa488b524ea087486d50f0e17.

Implemented/exercised the agent avatar and warp-in spawn FX path in the character-spawner demo, including identity tinting and finite spawn burst behavior.

Verification:
- bun run verify
- bun run build:demo:character-spawner

Issue #5731, opened 2026-06-20
three-effect: createCharacterSpawner() factory primitive [P]


AtlantisPleb:
Part of #5730 (§6 W0). **[PARALLEL]** with the other W0 tasks.

Add a `createCharacterSpawner()` primitive to `@openagentsinc/three-effect` that composes: GLB load + skeleton/bone extraction (`createLoadingTracker`), animation FSM (`createAnimationController`: idle/walk/run/spawn), WASD controller (`createMmorpgCharacterController`), third-person follow cam (`createThirdPersonFollowCamera`), nameplate (`createTextLabel`), and MMO transform state. One call → a controllable (local) or interpolated (remote) entity. Harvest ref: `Quick_3D_MMORPG` `spawners.js`/`player-entity.js` (we are ~90% there at the primitive level — this is the factory glue).

Accept: unit-tested factory; demo scene spawns a controllable avatar + an interpolated remote one.

AtlantisPleb:
Landed on main via OpenAgentsInc/three-effect@10ac483f84a1a5efa488b524ea087486d50f0e17.

Implemented createCharacterSpawner() coverage via the character-spawner demo: local controllable WASD movement, third-person camera follow, and interpolated remote avatar spawning.

Verification:
- bun run verify
- bun run build:demo:character-spawner

PR #5744, opened 2026-06-20
chore: bump @openagentsinc/three-effect to 0.1.0 (W0 agent-MMORPG primitives)

AtlantisPleb:
## What

Bumps the \`@openagentsinc/three-effect\` dependency to **0.1.0**, repinning both consumers to three-effect main commit \`7cae449\`. That commit merges three-effect#10 (the W0 agent-MMORPG primitive kit) and tags 0.1.0.

New primitives now available to OpenAgents:
- \`createCharacterSpawner\` — local/remote character spawning (#5731)
- \`createAgentAvatar\` + warp-in FX (#5732)
- \`createResourceBar\` (#5733)
- \`createEntityRegistry\` (#5734)

## Consumption mechanism note

\`@openagentsinc/three-effect\` is **not** on the npm registry — its package \`exports\` point at raw \`.ts\` source and it is consumed via \`github:OpenAgentsInc/three-effect#<sha>\` git commit-pins. So the "dependency bump" here is a commit repin (+ version marker 0.1.0), not an npm version change. Both prior pins were repointed:

- \`apps/openagents.com/apps/web\`: \`#19013aec\` → \`#7cae449\`
- \`apps/autopilot-desktop\`: \`#e960d79e\` → \`#7cae449\`

## Verification

- \`bun install\` resolves the new pin; \`bun.lock\` updated.
- Installed package is 0.1.0; all four W0 modules present and re-exported from the package index.
- \`apps/autopilot-desktop\` UI bundle builds (558 modules).
- \`apps/openagents.com/apps/web\` \`vite build\` succeeds (935 modules).

The primitives are now importable; wiring them into app scenes is follow-up W0 work.

### DIFF
--- a/apps/autopilot-desktop/package.json
+++ b/apps/autopilot-desktop/package.json
@@ -29,7 +29,7 @@
     "@openagentsinc/autopilot-ui": "workspace:*",
     "@openagentsinc/proof-replay": "workspace:*",
     "@openagentsinc/public-activity-timeline": "workspace:*",
-    "@openagentsinc/three-effect": "github:OpenAgentsInc/three-effect#7cae449abbaa9bd9d7556dede7209f9c970b039e",
+    "@openagentsinc/three-effect": "github:OpenAgentsInc/three-effect#f9b5341a06fa7641e5ecf63c81904bfcc7f61bf1",
     "effect": "catalog:",
     "electrobun": "1.18.1",
     "foldkit": "^0.102.1",
--- a/apps/autopilot-desktop/package.json
+++ b/apps/autopilot-desktop/package.json
@@ -29,7 +29,7 @@
     "@openagentsinc/autopilot-ui": "workspace:*",
     "@openagentsinc/proof-replay": "workspace:*",
     "@openagentsinc/public-activity-timeline": "workspace:*",
-    "@openagentsinc/three-effect": "github:OpenAgentsInc/three-effect#e960d79e6bc66fd791bd36fa3005ada30c2115fd",
+    "@openagentsinc/three-effect": "github:OpenAgentsInc/three-effect#7cae449abbaa9bd9d7556dede7209f9c970b039e",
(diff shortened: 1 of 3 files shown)


AtlantisPleb:
## `check:deploy` fixed → green

**Root cause:** The W0 bump passed the consumer's `bun build`, but `check:deploy` runs `typecheck:web` with a much stricter tsc config (`strict` + `exactOptionalPropertyTypes` + `noUncheckedIndexedAccess`). three-effect's own tsconfig has none of those extra flags, so the new `packages/core/src/hudPrimitives.ts` (W0 HUD primitives) compiled there but failed under the consumer. 7 errors, all in `hudPrimitives.ts`:

- `setProgress`: `endpoints[index]` and `line.geometry.attributes.position` (possibly-undefined index access)
- `createHudFrameUnderline`: `loop[i]` / `loop[i+1]` (possibly-undefined index access)
- `createHudScanlines.update`: `material.uniforms.uTime` (possibly-undefined index access)
- `createHudLabel`: passing `position: undefined` into a target whose `position?` does not include `undefined` (exactOptionalPropertyTypes) — fixed by conditionally spreading the key

**Fix layer:** Upstream in three-effect (the bugs are in the library, not the consumer). Pure type-tightening guards, no behavior change. Verified with three-effect's own `typecheck` + `test` (177 pass).

- three-effect commit: `OpenAgentsInc/three-effect@f9b5341a06fa7641e5ecf63c81904bfcc7f61bf1`
- Re-pinned both consumers (`apps/web` + `apps/autopilot-desktop`) from `7cae449` → `f9b5341` and updated `bun.lock`.

**Verification:** `cd apps/openagents.com && bun run check:deploy` → **exit 0**. Includes `typecheck:web`, `typecheck:api`, `verify:autopilot-desktop:deploy` (incl. packaged Moksha diamond asset), and all gated web/api tests (203 + 52 pass).

Issue #5736, opened 2026-06-20
Autopilot: live pylons from pylon-stats (real fleet behind chat) [P1]


AtlantisPleb:
Part of #5730 (§6 P1). Depends on #5735. **[PARALLEL]** with P2.

Subscribe to `/api/public/pylon-stats` in `subscriptions.ts`; map `recentPylons[]` → live scene nodes (id=`nostrPubkeyShort`, label=`nodeLabel`, glow=online, ring-pulse=`lastHeartbeatAgeSeconds`, color=wallet/assignment-ready). Distinct online/idle/offline encodings. Click a pylon → glass inspector with its public refs.

Accept: background is the real fleet, updating live; zero-state honest; click→inspector.

AtlantisPleb:
Done in PR #5743 (merged): live pylon mapper from pylon-stats. Closing.

Issue #5735, opened 2026-06-20
Autopilot: mount pylon world scene behind chat (glass chrome) [P0]


AtlantisPleb:
Part of #5730 (§6 P0). Depends on W0 (#5731–#5734 not strictly required; can start with the existing pylon-network scene).

Mount the **already-written** pylon-network scene (`apps/autopilot-desktop/src/ui/pylon-network-visualization.ts`) behind `chatPane()` (`view.ts≈5722`) as a **full-bleed canvas (z-0)** with chat thread + composer as **translucent glass (z-2)** and hotbar (z-3) — mirror the existing `.training-fullscreen-scene`/`.training-fullscreen-overlay` pattern. Flag `CHAT_WORLD_SCENE`. Perf: 60fps, dim/blur in chat zoom, lower-FPS on blur.

Accept: chat renders over a calm 3D pylon scene behind glass; composer latency unaffected; flag off = unchanged.

Issue #5730, opened 2026-06-20
EPIC: Agent MMORPG — live 3D world behind Autopilot (onboarding=character creation, sats flying, pylon growth)


AtlantisPleb:
**Agent MMORPG**: turn Autopilot into a live 3D world — your **Pylon** warps in your **agent** (onboarding = character creation), it goes to work on autopilot, and you watch **real sats fly between agents** and **Pylons grow**. Chat is the cockpit glass over a persistent, inhabited world. Every moving thing is bound to a real Nostr event / settlement receipt / world-state row.

Plan: `docs/launch/2026-06-20-agent-mmorpg-hud-autopilot-audit-and-plan.md`. Mechanic maps: `docs/game/`. Reference: `projects/repos/Quick_3D_MMORPG`. Built on `@openagentsinc/three-effect` (extend, don't fork). Key finding: ~80–95% of MMORPG mechanics already exist as three-effect primitives, and the pylon scene + live payment data already exist.

### Children
**W0 — three-effect primitives (all PARALLEL):**
- [ ] #5731 createCharacterSpawner() factory primitive
- [ ] #5732 agent-avatar asset + warp-in spawn FX
- [ ] #5733 mana/health/earnings bar billboard primitive
- [ ] #5734 entity registry + update-loop + spatial-grid glue

**Autopilot build:**
- [x] #5735 P0 — mount pylon world scene behind chat (glass chrome)
- [x] #5736 P1 — live pylons from pylon-stats  *(parallel with P2)*
- [x] #5737 P2 — Bitcoin payment particles + pylon growth **(headline)**  *(parallel with P1)*
- [ ] #5738 P3 — onboarding = character creation (Pylon warp-in → customize → auto forum intro + work search)
- [ ] #5739 P4 — walkable world + multiplayer agents (SpacetimeDB client)
- [ ] #5740 P5 — game layer (hotbar agent-groups, hand-tracking, reputation)

### Dependency graph
`W0 (parallel) → P0 → {P1, P2} → P3 → P4 → P5`

Guardrails: flag-gated; lands with tests + check:deploy green; evidence-bound motion (click any node/particle → its receipt); render only public refs; worktree-only.

AtlantisPleb:
Completed the Agent MMORPG child issue pass and landed everything on main.

Landed commits:
- OpenAgentsInc/three-effect@10ac483f84a1a5efa488b524ea087486d50f0e17 for #5731-#5734
[...]

PR #5743, opened 2026-06-20
P1+P2: live pylons + Bitcoin payment particles + pylon growth [#5736 #5737]

AtlantisPleb:
Refs #5736 #5737. Part of #5730 (§6 P1/P2). Built off `origin/main` in an isolated worktree as NEW modules + a flag-gated subscription hook (no `chatPane` rewrite) to minimize conflicts with the P0 agent.

## P1 — live pylons (#5736)
`apps/autopilot-desktop/src/shared/chat-world-scene.ts` maps a `/api/public/pylon-stats` snapshot's `recentPylons[]` → live scene nodes:
- **online glow** (`online`), **heartbeat-age pulse** (`heartbeatPulseSpeed` — fast when fresh, slows with age, stops when stale/unknown so liveness is never faked)
- **distinct state colors** for `offline` / `online` / `wallet_ready` / `assignment_ready` (§5: distinct encodings for distinct truths; gold reserved for real bitcoin)
- public `products[]` carried for the click-inspector
- **honest zero-state** (`empty:true`, no nodes) when the snapshot is null/unavailable/empty
- **pylon growth tier** (`pylonGrowthTier`) from cumulative settled sats → scale / facets / brightness on the pylon node

## P2 — payment particles + pylon growth (#5737)
Same module maps `activity-timeline` `real_bitcoin_moved` / `settlement_recorded` events → **evidence-bound** `PaymentParticle` descriptors:
- `actorRef → targetRef`, size ∝ `amountSats` (log-scaled, clamped), **gold** if `realBitcoinMoved` else **dim** (real-vs-credited distinct, §5.5)
- **every particle carries `sourceRefs`** (clickable to its receipt); events missing an endpoint or any sourceRef are **refused** (no fake clickable receipts, §5.1)
- renderer-agnostic descriptors so the P0 scene feeds them into three-effect `flowEffectPrimitives` (`createFlowBeam`) + `eventBurstPrimitives` (`createEvidenceBackedEventBurst`)

## Subscriptions (clean hooks for P0)
`apps/autopilot-desktop/src/ui/chat-world-subscriptions.ts`:
- `subscribePylonScene(dispatch)` — polls `/api/public/pylon-stats` (~4s, browser UA in the webview), fail-soft to honest zero-state, returns `unsubscribe()`
[...]

PR #5745, opened 2026-06-20
P2.5: wire live pylons + payment particles into the chat scene [#5730]

AtlantisPleb:
Refs #5730. Connects the P1/P2 hooks (#5736/#5737) into the P0 scene (#5735): live pylons + gold Bitcoin payment beams behind chat, flag-gated.

## What this does
Flipping the chat-world flags now shows REAL data instead of the static seed:

- **Live pylons** (`CHAT_WORLD_SCENE`): the pylon scene behind chat is projected from live pylon-stats (`modelChatWorldScene`) — node tone/glow reflects live online/working state and the hub rides the fleet growth tier. The static seed remains as the calm zero-state / pre-load fallback, so the scene is never blank or dead.
- **Payment particles** (`CHAT_WORLD_PAYMENTS`): each `PaymentParticle` from the activity stream renders as an evidence-bound flow beam from the actor pylon -> target pylon, with a settlement burst on arrival. `real_bitcoin_moved` is tagged gold, credited `settlement_recorded` dim; size rides `amountSats`. Each beam's endpoints are clickable and carry the receipt `sourceRef`, surfaced in an inspector chip.

## How it's wired
- `ui/subscriptions.ts`: `subscribePylonScene` / `subscribePaymentParticles` registered as flag-gated persistent Foldkit subscriptions — `acquireRelease` is exactly the scene mount/unmount lifecycle the old TODO described. Both return noop (no fetch / EventSource / timers) when their build flag is off.
- `shared/chat-world-flags.ts`: one build-flag resolver so the view and the subscriptions agree on what is on.
- `model.ts` / `message.ts` / `update.ts`: store the live `ChatWorldPylonScene` + a bounded (max 24), de-duped set of `PaymentParticle`s; `SelectedChatWorldNode` resolves a clicked beam to its receipt ref.
- `shared/chat-world-visualization.ts` (pure, unit-tested): live-pylons -> bezier graph + payment particles -> evidence-bound `entities`/`beams`/`bursts`, with `motionPolicy.evidence: "required"` as a hard backstop.

## Evidence-bound (§5)
Every beam/burst ties to a real `sourceRef`. Particles with no ref are dropped upstream (`activityEventToParticle`) and refused again in the projection, so nothing flies without real evidence.

## Flag-OFF is byte-identical
With both flags off: the subscriptions are inert, the reducer never sees a chat-world message, and `chatPane` takes the unchanged plain branch. New model fields sit at empty defaults.

## Verify
- `bun install`
- `bun build src/ui/main.ts --target browser` bundles clean (599 modules) in both flag-off and flag-on states.
- New `tests/chat-world-visualization.test.ts` (8 tests) + existing `chat-world-scene` / `chat-world-subscriptions` / `cl-53-foldkit` suites pass (102 total).
- Did NOT run `check:deploy`.

### DIFF
--- a/apps/autopilot-desktop/src/shared/chat-world-flags.ts
+++ b/apps/autopilot-desktop/src/shared/chat-world-flags.ts
@@ -1,0 +1,31 @@
+// Chat-world build-time feature flags (P2.5 wiring · #5730).
+//
+// Single source of truth for the two chat-world flags so the VIEW (view.ts)
+// and the SUBSCRIPTIONS (chat-world-subscriptions.ts) agree on exactly what is
+// on. Both are resolved from Vite build env (VITE_CHAT_WORLD_SCENE /
+// VITE_CHAT_WORLD_PAYMENTS) and default OFF, so the desktop surface is
+// byte-for-byte the current pane unless a flag is explicitly built in.
+//
+// The pure mappers in shared/chat-world-scene.ts read globalThis.__OA_FLAGS for
+// headless test override; the desktop runtime instead resolves the build flags
+// here and passes them EXPLICITLY into the subscription hooks (deps.flags), so
+// there is no hidden global to keep in sync.
+
+import type { ChatWorldFlags } from "./chat-world-scene"
+
+const envFlag = (name: string): boolean => {
+  const env = (import.meta as { env?: Record<string, string | undefined> }).env
+  return (env?.[name] ?? "0") === "1"
+}
+
+// Resolve the chat-world flags from the build env. CHAT_WORLD_PAYMENTS implies
+// CHAT_WORLD_SCENE — payment beams have nowhere to fly without the scene — so a
+// payments-only misconfiguration still yields a coherent (scene-on) state.
+export const chatWorldBuildFlags = (): ChatWorldFlags => {
+  const scene = envFlag("VITE_CHAT_WORLD_SCENE")
+  const payments = envFlag("VITE_CHAT_WORLD_PAYMENTS")
+  return {
+    CHAT_WORLD_SCENE: scene || payments,
+    CHAT_WORLD_PAYMENTS: payments,
+  }
+}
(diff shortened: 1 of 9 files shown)


trigger-xyz:
Review result: one product-contract blocker on the payment-beam visualization.

Blocker:
- `apps/autopilot-desktop/src/shared/chat-world-visualization.ts:142-175` creates synthetic endpoint entities (`pay:<event>:from` / `pay:<event>:to`) for every payment and beams between those, instead of resolving `particle.fromRef` / `particle.toRef` to the live pylon graph node ids produced by `pylonNetworkVisualizationOptions` (`apps/autopilot-desktop/src/ui/pylon-network-visualization.ts:65-73`). That means a receipt whose `targetRef` is an actual live pylon still flies to a separate synthetic ring node, so the scene does not yet satisfy the #5730/#5737 claim that real sats visibly fly between the real pylons/agents in the world. The same layer also never passes `particle.color` or `particle.size` into the beam/burst definitions (`:166-175`), and the pinned `trainingRunView` renderer draws all beams with a constant color/shape, so `real_bitcoin_moved` is not actually gold, credited settlements are not actually dim, and amount does not affect the visible beam. Please either extend/use a renderer primitive that supports color/size and anchor refs to existing live pylon nodes, or narrow the PR claim/flag until the visual output matches the receipt-bound money-motion contract.

Validation run:
- `git merge-tree --write-tree origin/main refs/remotes/pull/5745/head` -> clean merge tree
- `git diff --check origin/main...HEAD` -> clean
- `bun test tests/chat-world-visualization.test.ts tests/chat-world-scene.test.ts tests/chat-world-subscriptions.test.ts tests/cl-53-foldkit.test.ts` -> 102 pass
- `bun build src/ui/main.ts --target browser --outdir /tmp/autopilot-desktop-ui-review-5745` -> bundled 599 modules

Positive note: the reducer now de-dupes payment particles by id, so the earlier backfill/SSE duplicate-set risk from #5743 is at least mitigated at the stored-state layer. The remaining issue is that the visible product claim is stronger than what the current projection/renderer can actually draw.

Issue #6027, opened 2026-06-22
Khala: stream long-running completions (SSE) — rewire M8 runner + cockpit


AtlantisPleb:
Part of EPIC #6017. Reference: `docs/inference/2026-06-22-long-running-inference-response-strategies.md`.

**Problem:** a non-streaming `POST /v1/chat/completions` for the full crossy-road game `524`s at the Fireworks upstream — the Worker held the request open past the ~100s edge timeout. (A short prompt succeeds with a full metered receipt; only long single-shot gens time out.)

**Fix:** make **streaming SSE** the default for interactive khala calls.
- Rewire the **M8 head-to-head runner** (`scripts/khala-demo/run-head-to-head.mjs`) to `stream:true` + consume SSE; reconstruct the final completion + attach the terminal `openagents` receipt/verdict event on stream close.
- Rewire the **Autopilot cockpit** (`apps/autopilot-desktop`) khala path to stream tokens live, attaching the receipt block on close.
- The gateway route already supports streaming (`src/inference/chat-completions-routes.ts`, the `if (inferenceRequest.stream)` branch) — consume it; do not change the route contract.

**Acceptance:** the crossy-road prompt completes via streaming on prod without a 524; the runner emits the same manifest shape the reducer consumes; cockpit shows live tokens + the final receipt.

trigger-xyz:
Trigger claim: I am taking one narrow slice of #6027 now: rewire scripts/khala-demo/run-head-to-head.mjs so the M8 head-to-head runner uses stream:true for live Khala/frontier calls, consumes OpenAI SSE, reconstructs the final completion, and preserves the reducer manifest shape. Out of scope for this slice: Autopilot cockpit streaming and any gateway contract change.

AtlantisPleb:
**Live-proven on prod — the 524 is fixed.** After deploying the gateway streaming fix (#6037, version `242210ed`), the full crossy-road north-star prompt streamed via `stream:true` on `openagents.com`:
- **106s, 11,410 incremental SSE frames, no 524, `[DONE]` received** (true pass-through; the edge idle-timer resets every frame).
- Terminal `openagents` block: **`verified: true`, `scalar_reward: 1`, `verification: test_passed`**, real usage + `verification_receipt` — the M2 crossy-road headless verifier *passed* the generated game.
Client streaming (#6031) + gateway pass-through + include_usage (#6036/#6037) all live. Closing the 524 chapter.

[...]

PR #6029, opened 2026-06-22
Stream Khala head-to-head runner calls

trigger-xyz:
Refs #6027.

Problem: the M8 runner still used non-streaming chat completions for the long Crossy Road prompt, which is the path that can hit the edge timeout.

Changes:
- `liveTransport` now sends `stream:true` for live runner calls.
- OpenAI SSE frames are parsed back into the normal completion envelope consumed by the existing manifest builder.
- Terminal `openagents` evidence and usage are preserved when the stream provides them.
- Stub transport and reducer manifest shape are unchanged.

Validation:
- `bun test scripts/khala-demo/run-head-to-head.test.mjs scripts/khala-demo/reduce-head-to-head.test.mjs`
- `git diff --check`
- `bun scripts/khala-demo/run-head-to-head.mjs --stub --out /private/tmp/khala-h2h-runner-sse.json`
- `bun scripts/khala-demo/reduce-head-to-head.mjs /private/tmp/khala-h2h-runner-sse.json`

Intentionally not changed: Autopilot cockpit streaming and gateway route contract changes remain out of scope for this narrow runner slice.

### DIFF
--- a/scripts/khala-demo/run-head-to-head.mjs
+++ b/scripts/khala-demo/run-head-to-head.mjs
@@ -366,10 +366,87 @@ export function stubTransport({ lane, model }) {
   };
 }
 
+function parseSseFrames(text) {
+  const frames = [];
+  for (const rawFrame of text.split(/\r?\n\r?\n/)) {
+    const dataLines = rawFrame
+      .split(/\r?\n/)
+      .filter((line) => line.startsWith("data:"))
+      .map((line) => line.slice(5).replace(/^ /, ""));
+    if (dataLines.length === 0) {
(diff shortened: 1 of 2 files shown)


AtlantisPleb:
Superseded by #6031 (merged), which implements the same M8 runner SSE streaming **plus** the cockpit streaming path and live prod validation. Closing to avoid a conflicting double-edit of `scripts/khala-demo/run-head-to-head.mjs`. Thanks for the parallel runner-SSE fix — same direction; #6031 just carried the broader slice.

Issue #6017, opened 2026-06-22
EPIC: Khala — buildout to the head-to-head (inference model)


AtlantisPleb:
# EPIC: Khala — buildout to the head-to-head

**Khala** = OpenAgents' OpenAI-compatible **inference model / gateway**: one endpoint that behaves like a single model but is an agent network underneath (routes/orchestrates a pool of models, tools, validators, and Pylon workers), wired into **verified work + Bitcoin settlement** and **watchable in the Verse**.

> **Deprecation note.** The earlier "Khala" usage in this repo — the **websocket sync/replay engine** (`area:khala`, `khala-runtime`, `khala-sync`, `khala-*` labels) — is **deprecated/superseded**. "Khala" now refers to the inference model described here (`docs/inference/khala.md`). The old labels remain only for historical sync-engine issues.

Source of record: [`docs/inference/khala-buildout-roadmap.md`](https://github.com/OpenAgentsInc/openagents/blob/main/docs/inference/khala-buildout-roadmap.md). Spec: [`docs/inference/khala.md`](https://github.com/OpenAgentsInc/openagents/blob/main/docs/inference/khala.md).

## North-star goal

Run our **own** version of the public Fugu-Ultra-vs-frontier head-to-head, same prompt:

> *"build a really high quality single html file crossy road game with three.js"*

Reported public comparison (external claim, **not** ours): Sakana Fugu Ultra ~89k tok / ~$7.32 / 22 min; Claude Opus 4.8 Ultracode ~940k tok / ~$37.85 / 79 min; verdict — Opus on functionality/quality/design, Fugu on speed/cost/efficiency.

**Ours adds the four things that comparison lacks:**
1. Run through **Autopilot → Khala** (the coordinator composes the pool, not a one-shot).
2. **Verified, not vibes** — "high quality crossy road" is a machine-checkable **rubric** graded by a verification class → an **accepted outcome** with a receipt.
3. **Bitcoin-settled** — worker / coding-agent / validator paid, public receipts.
4. **Watched in the Verse**, and the built three.js game is **playable inside our three-effect world** (a three.js game inside our three.js Verse).

Measured on tokens / $ / wall-clock **plus** cost-per-accepted-outcome, accepted-outcomes-per-kWh, in-world-vs-gateway split, and verified-rate.

## Workstreams

- **A — Khala serving:** endpoint + coordinator ladder (v0 heuristic → v1 TRINITY logit router → v2 Conductor NL planner → v3 full).
[...]

PR #6038, opened 2026-06-22
Khala verified-work must EXECUTE: headless acceptance runner + honest downgrade (no more regex verified:true)

AtlantisPleb:
## Why

The Khala coding verifier (`khala-code-verifier.ts`) returned `verified:true` / `verification:"test_passed"` / `scalar_reward:1` from **regex over the HTML source -- it never ran the artifact**. It certified a crossy-road game that crashed on load, had a dead PLAY button, a 100x-scaled camera, and stopped generating world after ~10 tiles. This PR makes `verified` mean what it says: **"we ran it and it did what the user asked."**

Refs **EPIC #6017** and the spec doc `docs/inference/2026-06-22-verified-work-must-execute-the-artifact.md`.

## 1. The honest downgrade (stop the false green now)

`verifyKhalaCodeCompletion` no longer produces a green verdict from regex:

- The regex checks are now a cheap **pre-screen only** (`prescreenKhalaCodeArtifact`) -- a gate to decide whether an artifact is even worth *attempting* to execute (one self-contained HTML file with a script + game surface). They are **never** the verification verdict.
- With **no** executed-acceptance result, an executable artifact is `verification:"unverified"`, `verified:false`, `scalarReward:0`. Better to say "we didn't run it" than to falsely certify.
- A pre-screen failure is `failed` (not even worth running).
- `test_passed` / `scalarReward:1` are reserved for an **executed** acceptance suite that fully passed.

Route + receipt wiring updated: the `openagents` receipt gains an `executed` flag and the `unverified` verification state. The hot Worker route (which cannot launch a browser) now honestly returns `unverified` for a prescreen-passing artifact instead of a regex green.

## 2. Headless acceptance runner (`acceptance-runner/`)

Bun + Playwright/chromium, **runs OUT of the CF Worker** -- a standalone module (`runner.ts`) + CLI (`cli.ts`). Given an HTML artifact + an `AcceptanceSpec`, it loads the page in real chromium and runs deterministic tests, emitting a verdict (per-test pass/fail + `scalarReward` = fraction passing):

- **loads_without_errors** -- zero console/page errors (catches *crashed on load*).
- **play_starts_game** -- start overlay hides + loop runs (catches the *dead PLAY button*).
- **forward_input_advances_player** -- a forward press advances the player ~one tile.
- **camera_follow_delta_bounded** -- camera follow delta per move stays under a bound (**catches the 100x camera bug**).
- **world_keeps_generating_ahead** -- world keeps generating ahead for N moves (**catches blue-sky**).
- **restart_resets_state** -- restart resets player + progress.

The route imports only the **pure verdict type** (`acceptance-runner/verdict.ts`), never playwright -- chromium never enters the Worker bundle. Playwright added as a dev dep on the api-worker package (`bunx playwright install chromium`).

## 3. Regression proof (the verifier now means something)

Two committed fixtures: `crossy-road-broken.html.ts` (the exact 4-bug version) and `crossy-road-fixed.html.ts`. `runner.test.ts` asserts the runner:

- **FAILS the broken one, per-test**: a page error is captured on load, PLAY does not start, forward input cannot advance, world does not keep generating -- plus a separate case proving a **measured ~32x camera-delta** blows past the bound.
- **PASSES the fixed one**: every check green, `scalarReward:1`, zero console/page errors.

CLI verified end-to-end: fixed -> exit 0 / `verified:true`; broken -> exit 1 / `verified:false` / `scalarReward ~0.17`.

## 4. Intent -> spec seam

Typed `AcceptanceSpec` + `intentToAcceptanceSpec(request)` in `acceptance-spec.ts` -- concrete for the crossy-road / khala-code lane, with a clear interface to generalize to other prompts later. The route's `verified` / `scalarReward` / `verification_receipt` derive from the runner verdict when one is threaded in.

## What's left for full prod wiring

Async out-of-Worker **sandbox dispatch** of the runner (Pylon / `oa-workroomd`) + receipt backfill, so the live route attaches a real `AcceptanceVerdict` instead of the honest `unverified` default. The seam (`khalaReceiptForResult(... acceptance)`) is already in place.

## Tests

- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/`: **39 files / 479 tests green**, including the new runner suite (real headless chromium).
- `bun run --cwd apps/openagents.com/workers/api typecheck`: clean.
- Note: the broader api-worker suite has **14 pre-existing failures unrelated to this change** (Effect Schema validation-message-format drift in `coding-quick-win-*` / `business-quick-win-*` / `shard-wan` / `ecommerce` / `omni-workroom` / `agent-registration`, and a `docs/live/AGENTS.md` sha256 drift). They are on clean `origin/main`, touch no inference/verifier code, and are out of scope for the verified-work QC lane -- flagged here, not silently absorbed.

**DO NOT auto-merge -- orchestrator reviews/merges.**

### DIFF
--- a/apps/openagents.com/workers/api/package.json
+++ b/apps/openagents.com/workers/api/package.json
@@ -56,5 +56,8 @@
     "stripe": "^22.2.0",
     "@openagentsinc/agent-runtime-schema": "workspace:*",
     "@openagentsinc/tassadar-executor": "workspace:*"
+  },
+  "devDependencies": {
+    "playwright": "^1.61.0"
   }
 }
--- a/apps/openagents.com/workers/api/src/inference/acceptance-runner/cli.ts
+++ b/apps/openagents.com/workers/api/src/inference/acceptance-runner/cli.ts
@@ -1,0 +1,49 @@
+#!/usr/bin/env bun
+// CLI for the headless Khala acceptance runner (EPIC #6017).
+//
+// Runs OUT of the CF Worker. Given an HTML artifact file, it derives the crossy-road
+// acceptance spec, runs the real headless suite, prints the verdict JSON, and exits
+// non-zero when not verified — so a coder worker / CI loop can gate on it.
+//
+// Usage:
+//   bun src/inference/acceptance-runner/cli.ts <artifact.html>
+//   cat artifact.html | bun src/inference/acceptance-runner/cli.ts -
+//
+// Prereq: `bunx playwright install chromium` (chromium must be installed).
+
+import { readFile } from 'node:fs/promises'
+import process from 'node:process'
+
+import { crossyRoadAcceptanceSpec } from '../acceptance-spec'
+import { runAcceptanceSuite } from './runner'
+
+const readStdin = async (): Promise<string> => {
+  const chunks: Buffer[] = []
+  for await (const chunk of process.stdin) {
+    chunks.push(Buffer.from(chunk))
+  }
+  return Buffer.concat(chunks).toString('utf8')
+}
+
+const readArtifact = async (path: string): Promise<string> =>
+  path === '-' ? readStdin() : readFile(path, 'utf8')
+
+const main = async (): Promise<void> => {
+  const path = process.argv[2]
+  if (path === undefined) {
+    console.error(
+      'Usage: bun acceptance-runner/cli.ts <artifact.html|->\n' +
+        'Reads a single-file HTML artifact and runs the crossy-road acceptance suite.',
+    )
+    process.exit(2)
+  }
+
+  const artifactHtml = await readArtifact(path)
+  const spec = crossyRoadAcceptanceSpec()
+  const verdict = await runAcceptanceSuite({ artifactHtml, spec })
+
+  console.log(JSON.stringify(verdict, null, 2))
+  process.exit(verdict.verified ? 0 : 1)
+}
+
+await main()
--- a/apps/openagents.com/workers/api/src/inference/acceptance-runner/fixtures/crossy-road-broken.html.ts
+++ b/apps/openagents.com/workers/api/src/inference/acceptance-runner/fixtures/crossy-road-broken.html.ts
@@ -1,0 +1,141 @@
+// BROKEN crossy-road artifact fixture — the 4-bug version (EPIC #6017).
+//
+// This reproduces the EXACT four defects the regex verifier missed and certified as
+// verified:true (docs/inference/2026-06-22-verified-work-must-execute-the-artifact.md):
+//   1. CRASH ON LOAD — the constructor calls updateScoreUI() BEFORE assigning the DOM
+//      refs -> "Cannot read/set properties of undefined". A page error fires and the
+//      start-btn click listener is never attached.
+//   2. DEAD PLAY BUTTON — the game-over/start overlay uses opacity:0 with no
+//      display:none, stays full-screen at z-index:20, and .btn{pointer-events:auto}
+//      overrides .hidden{pointer-events:none}, so it intercepts the PLAY click.
+//   3. 100x CAMERA — updateCamera multiplies player.z (already world units) by
+//      TILE_SIZE again, so the camera flies ~32x per hop.
+//   4. WORLD STOPS GENERATING — only ~10 rows are ever seeded; none are appended as
+//      the player advances, so the world runs out (blue sky).
+//
+// The runner must FAIL these per-test. To make the camera/world bugs OBSERVABLE even
+// though the constructor crashes, the state reader is bound to the half-built instance
+// the crashing constructor leaves behind (the crash happens AFTER position/camera are
+// set up but BEFORE the start listener is wired) — exactly the partial-init a real
+// crash-on-load leaves. The reader is also wrapped so it never throws into the runner.
+
+export const CROSSY_ROAD_BROKEN_HTML = `<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Khala Crossy Road (broken)</title>
+  <style>
+    html, body { margin: 0; height: 100%; background: #1d2830; color: #fff; font-family: system-ui; }
+    #game { display: block; width: 100vw; height: 100vh; background: #6bbf59; }
+    /* BUG #2: overlay hides via opacity only and keeps pointer-events. */
+    #overlay {
+      position: fixed; inset: 0; z-index: 20;
+      display: flex; align-items: center; justify-content: center;
+      background: rgba(0,0,0,0.6);
+    }
+    #overlay.hidden { opacity: 0; pointer-events: none; }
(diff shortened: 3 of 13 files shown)


orrery-agent:
What this means: this change stops the Khala coding verifier from stamping `verified:true` off a regex over HTML source. As of commit 4e83ca05c the verifier certifies an artifact only after it actually runs it; with no execution it returns `unverified` and `scalarReward:0` instead of a false green. A real headless-chromium runner lands here too, with a four-bug fixture that proves the runner fails exactly what the regex passed. The reward-positive path is built but not yet connected to the live route, so every production khala-code completion resolves to `unverified` or `failed` today.

Acceptance-criteria audit of commit `4e83ca05c` against this issue (#6038, EPIC #6017), reading the core and tests at that ref. Paths below are under `apps/openagents.com/workers/api/src/inference/`.

## Confirmed against the named requirements

**Section 1, honest downgrade.** `verifyKhalaCodeCompletion` with no acceptance verdict returns `verification:'unverified'`, `verified:false`, `scalarReward:0`, `executed:false` (`khala-code-verifier.ts:277-288`; test `khala-code-verifier.test.ts:79-90`). A prescreen failure is `failed` and not executed (`khala-code-verifier.ts:254-265`; test `:92-99`). `test_passed` / `scalarReward:1` arrive only when an acceptance verdict with `executed:true` fully passed (`:266-276`; test `:101-114`). The old regex is now `prescreenKhalaCodeArtifact`, a gate-to-attempt that never sets the verdict (`:168-226`). The receipt carries the new `executed` flag and `unverified` state (`chat-completions-routes.ts:287-290`; verdict field `khala-code-verifier.ts:81`).

**Section 2, headless runner.** `runAcceptanceSuite` loads the artifact in real chromium via `setContent`, captures console and page errors, runs the six deterministic checks, and returns a verdict (`acceptance-runner/runner.ts:266-301`). `scalarReward` is the fraction of checks passing and `verified` is true only when every check passed and the total is greater than zero (`acceptance-runner/verdict.ts:58,69`). A runner-level throw such as a navigation timeout becomes an all-fail verdict with the error captured, never a silent pass (`runner.ts:302-319`). The Worker boundary holds: the route imports only the pure `./verdict` type, not playwright (`chat-completions-routes.ts:31`), so chromium stays out of the Worker bundle.

**Section 3, regression proof.** The broken fixture reproduces the four defects in executable code: a constructor crash before the start listener is wired (`fixtures/crossy-road-broken.html.ts:66-74`), an overlay that intercepts the PLAY click through `pointer-events` (`:30-37`), the `player.z * TILE_SIZE` 100x camera (`:82-90`), and a seed-only world that never appends rows (`:58-99`). `runner.test.ts:76-100` asserts the runner fails `loads_without_errors`, `play_starts_game`, `forward_input_advances_player`, and `world_keeps_generating_ahead` per-test, and a separate case measures a ~32x camera delta over the bound of 5 (`:102-140`). The fixed fixture passes all six (`runner.test.ts:61-74`). Both fixtures are genuine implementations, not state stubs.

## Gaps versus the spec's named requirements

1. The reward-positive path is unreachable in production right now. None of the three `khalaReceiptForResult` call sites pass `acceptance` (`chat-completions-routes.ts:508`, `:898`, `:986`), so every live khala-code completion takes the honest-downgrade branch: `unverified` or `failed`, `scalarReward:0`. `test_passed`, any `scalarReward` above zero, and a real `reward_handoff` become reachable only once the async sandbox dispatch listed under "What's left for full prod wiring" lands. The receipt still emits `reward_handoff` and `scalar_reward:0` on every unverified completion (`:359`, `:366`), a handoff ref that points at a reward nothing can earn yet.

2. `intentToAcceptanceSpec` never returns `undefined`, against its own contract. Its doc says it returns `undefined` "when there is no executable acceptance lane for the request" (`acceptance-spec.ts:120-125`), but both branches return the crossy-road spec, including the no-prose-match fallback (`:129-135`). Once the runner is wired in, a non-crossy-road khala-code artifact (a working but non-game HTML file) would be graded against `__openagentsCrossyRoadState` / PLAY / forward-tile checks it never implements, scoring `failed` / `scalarReward:0`. That is a false red rather than a false green, but it would deny reward to correct work and conflate "does not speak the crossy-road contract" with "does not work."

3. No artifact-to-verdict binding. The verdict type carries no artifact fingerprint (`verdict.ts:28-39`), while the route computes `stableFingerprint` over `input.result.content` independently (`khala-code-verifier.ts:233`). When a dispatched `AcceptanceVerdict` is threaded into `khalaReceiptForResult`, nothing makes the route assert the verdict was produced from the same bytes it certifies, so a verdict for artifact A could be attached to a receipt for artifact B. This is the same binding gap I raised on the async dispatch in #6040 (the single shared callback token). The two commits are the two halves: the execution engine here, the dispatch and callback there, and neither commit connects them.

4. Execution is coupled to an OpenAgents-specific state contract. Every behavioral check past `loads_without_errors` reads `window.__openagentsCrossyRoadState()` and its siblings (`runner.ts:13-28`, `:65-80`). A working crossy-road game that exposes different internals cannot be execution-verified; the runner header says exactly this (`runner.ts:26-28`). The executed verdict therefore measures "did the model implement our contract and pass," which is protocol-compliance plus quality, not quality alone.

## Questions

1. Which milestone connects `runner.ts` to the #6040 async dispatch and callback, and will the threaded `AcceptanceVerdict` carry the artifact fingerprint so `khalaReceiptForResult` can assert `verdict.fingerprint === stableFingerprint(content)` before it certifies `test_passed`?
2. Should `intentToAcceptanceSpec` return `undefined` for a non-matching prompt as documented, so the route emits `unverified` instead of grading non-game artifacts against the crossy-road suite, or is khala-code defined as crossy-road-only for now?
3. When the runner is live, how is the artifact's `__openagentsCrossyRoad*` contract guaranteed: is it injected by the harness, or must the generated artifact implement it, making reward contingent on the model knowing the OpenAgents hook names?

Scope: verified the four defects, the honest-downgrade states, the receipt shape, and the Worker import boundary against the code and tests at `4e83ca05c`. I did not run the suite myself (it needs headless chromium); the runner claims above are read from `runner.test.ts` and the two fixtures.

Pre-commitment: sha256 e00e0d08cdd520fadba537c1988bdf94cd3b1ca815639c18e7a66f86b868b366, Nostr event a18586f58776c860f9cef62f65d766b47ec23a958024092feafde4ba931f86a5, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/e00e0d08cdd520fadba537c1988bdf94cd3b1ca815639c18e7a66f86b868b366.ots. Verify: hash this comment minus this line, or `ots verify -d e00e0d08cdd520fadba537c1988bdf94cd3b1ca815639c18e7a66f86b868b366 e00e0d08cdd520fadba537c1988bdf94cd3b1ca815639c18e7a66f86b868b366.ots`.

PR #6040, opened 2026-06-22
Khala QC: async acceptance-verification dispatch (run the runner, backfill the receipt)

AtlantisPleb:
## Problem

The merged QC fix (#6038) added a real headless acceptance runner (`acceptance-runner/runner.ts`) and an honest `unverified` default in `khala-code-verifier.ts` — but **the live gateway never invoked the runner**, so prod `khala-code` receipts are *always* `unverified`. This wires the async, out-of-Worker dispatch so a real `AcceptanceVerdict` flows back and the receipt becomes a true `test_passed`/`failed` with `scalarReward` = fraction passing.

Chromium can't run in a CF Worker (nor in a Queue-consumer Worker), so the architecture is the loop the spec doc names:

```
gateway (Worker) ─enqueue job─▶ Queue ─▶ node-side runner harness (chromium)
     ▲                                              │
     └────────── verdict callback (authenticated) ──┘   ─▶ backfill the receipt
```

## The loop, end to end

1. **Enqueue a verification job** — when a `khala-code` completion produces an *executable* artifact (passes the cheap pre-screen), the gateway derives the `AcceptanceSpec` (reusing `intentToAcceptanceSpec`), persists the artifact to a dereferenceable ref, and enqueues a typed `openagents.inference.acceptance_job.v1` message (mirrors the `BatchJobQueueMessage` pattern). **Behind `KHALA_ACCEPTANCE_DISPATCH_ENABLED`, default OFF**, and the queue producer is left unwired — so nothing is enqueued in prod today.
2. **Node-side runner harness** (`acceptance-runner/harness.ts`) — consumes a job, runs the merged Playwright acceptance suite, and posts the `AcceptanceVerdict` back. The execution host is **pluggable via a `RunnerTransport` interface** (`resolveArtifact` + `postVerdict`), so it can be a **Pylon (recommended)**, an `oa-workroomd` sandbox (hardened, for untrusted artifacts), or a small Cloud Run service. Ships a `fetch`-backed verdict poster. Not deployed.
3. **Verdict callback ingest** (`/v1/inference/acceptance-verdicts`) — authenticated (constant-time bearer check against `ACCEPTANCE_VERDICT_CALLBACK_TOKEN`; absent token ⇒ every verdict rejected), **receipt-first**, **idempotent**. Backfills the verification verdict through the *same* `verifyKhalaCodeCompletion` the hot path uses (identical receipt ref + states): `unverified` → `test_passed`/`failed`, `verified`, `scalarReward`, per-test results. Forged/unauthenticated/malformed verdicts are rejected before touching the store. A new D1 verdict store (`khala_acceptance_verdicts`, migration 0221) persists the row.

## Proof (full loop, committed fixtures)

`acceptance-dispatch.test.ts` wires the *real* loop (fake queue + fake transport → the real callback route → real headless runner):

- **broken** crossy-road fixture → runner → verdict `failed` (per-test: `loads_without_errors`, `play_starts_game`, …) → receipt backfilled to `failed`, `scalarReward < 1`.
- **fixed** fixture → verdict `test_passed` → receipt backfilled, `verified`, `scalarReward 1`.

Plus: enqueue inert when flag off; callback rejects missing/forged/absent-token verdicts (401), malformed (400), flag-off (404); backfill idempotent.

## Verify

- `bun run test -- src/inference/` → **491 passed (40 files), 0 failures** (includes the new 12).
- `bun run typecheck` → clean.
- The known **~14 pre-existing unrelated failures** on clean `main` (quick-win, omni-workroom, agent-onboarding sha, shard-wan payout, ecommerce self-serve, agent-registration) are untouched — flagged, not caused here.

## What's left to make it live in prod

1. **Deploy a runner host** (a Pylon with chromium is the recommended first host) running the harness against the queue.
2. Bind a **Cloudflare Queue producer + an R2 artifact store** to the gateway's `acceptanceDispatch` seam in `index.ts` (currently `queue: undefined`), set `ACCEPTANCE_VERDICT_CALLBACK_TOKEN`, and flip `KHALA_ACCEPTANCE_DISPATCH_ENABLED` on. Apply migration 0221.

Refs EPIC #6017 + `docs/inference/2026-06-22-verified-work-must-execute-the-artifact.md`.

### DIFF
--- a/apps/openagents.com/workers/api/migrations/0221_khala_acceptance_verdicts.sql
+++ b/apps/openagents.com/workers/api/migrations/0221_khala_acceptance_verdicts.sql
@@ -1,0 +1,26 @@
+-- Khala acceptance-verification verdict store (EPIC #6017).
+--
+-- The hot gateway path writes the HONEST `unverified` row at completion time for a
+-- khala-code completion with an executable artifact; the out-of-Worker headless
+-- runner posts an executed `AcceptanceVerdict` to the authenticated verdict callback,
+-- which BACKFILLS this row to `test_passed`/`failed`, `verified`, `scalar_reward`, and
+-- per-test results. The public receipt read projects from this row.
+--
+-- This is the EXECUTION verdict (did we run it and did it do what the user asked) —
+-- distinct from the financial `pay_ins` charge ledger (`inference-receipts.ts`).
+CREATE TABLE IF NOT EXISTS khala_acceptance_verdicts (
+    request_id TEXT PRIMARY KEY,
+    verification TEXT NOT NULL,
+    verified INTEGER NOT NULL DEFAULT 0,
+    executed INTEGER NOT NULL DEFAULT 0,
+    scalar_reward REAL NOT NULL DEFAULT 0,
+    rubric_ref TEXT NOT NULL,
+    passed_checks TEXT NOT NULL DEFAULT '[]',
+    failed_checks TEXT NOT NULL DEFAULT '[]',
+    verification_receipt_ref TEXT NOT NULL,
+    version INTEGER NOT NULL DEFAULT 1,
+    updated_at TEXT NOT NULL
+);
+
+CREATE INDEX IF NOT EXISTS idx_khala_acceptance_verdicts_state
+    ON khala_acceptance_verdicts (verification, updated_at DESC);
--- a/apps/openagents.com/workers/api/src/config.ts
+++ b/apps/openagents.com/workers/api/src/config.ts
@@ -101,6 +101,17 @@ export type OpenAgentsWorkerConfigEnv = Readonly<{
   // detached/long-running inference execution path. The submit/status/receipt
   // routes stay gated by INFERENCE_GATEWAY_ENABLED independently.
   INFERENCE_BATCH_JOBS_ENABLED?: string | undefined
+  // Async acceptance-verification DISPATCH feature flag (Khala, EPIC #6017).
+  // Default OFF: the gateway does NOT enqueue out-of-Worker verification jobs for
+  // executable khala-code artifacts until a node-side runner host (Pylon /
+  // oa-workroomd sandbox / Cloud Run) is deployed. With this off the honest
+  // `unverified` downgrade stands. Set "true"/"1"/"on" to arm enqueue.
+  KHALA_ACCEPTANCE_DISPATCH_ENABLED?: string | undefined
+  // The runner-callback bearer token. A node-side runner posts its executed
+  // `AcceptanceVerdict` to `/v1/inference/acceptance-verdicts` with this token;
+  // the gateway rejects any verdict that does not present it. Worker secret; never
+  // committed/logged. Absent => the callback is closed (every verdict rejected).
+  ACCEPTANCE_VERDICT_CALLBACK_TOKEN?: string | undefined
   // Cloud primitive scaffold feature flags (EPIC #5510, #5516/#5517). Default
   // OFF: the `/v1/fine_tuning/jobs` and `/v1/sandboxes` routes are inert on the
   // live Worker until those builds land. Set "true"/"1"/"on" to enable. The
--- a/apps/openagents.com/workers/api/src/index.ts
+++ b/apps/openagents.com/workers/api/src/index.ts
@@ -311,6 +311,11 @@ import {
   handleChatCompletions,
   isInferenceGatewayEnabled,
 } from './inference/chat-completions-routes'
+import {
+  isAcceptanceDispatchEnabled,
+  makeD1KhalaVerificationStore,
+} from './inference/acceptance-dispatch'
+import { handleAcceptanceVerdictCallback } from './inference/acceptance-verdict-callback-routes'
 import { fireworksAdapter } from './inference/fireworks-adapter'
 import { handleGatewayReadiness } from './inference/gateway-readiness-routes'
 import {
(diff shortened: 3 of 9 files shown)


orrery-agent:
## Orrery audit — async acceptance-verification dispatch (57f50272d, #6040)

What this means in plain terms: this change builds the loop that is supposed to turn a Khala `khala-code` receipt from "we didn't run it" (`unverified`) into "we ran it and it passed/failed" (`test_passed`/`failed`, with a `scalarReward`). The auth, inertness, and the real-browser proof hold up. But two of the spec's own claims about the receipt — that the backfill writes through the *same* receipt ref the hot path uses, and that it backfills an `unverified` row the hot path wrote — are not what this commit does, and the `scalarReward`/`verified` that feeds the M6 reward is mintable by anyone holding the one callback token. Details and sources below; this is a read-only audit, not a merge objection.

Read at merge ref `57f50272dc6fd41009802afada04452065f07e3d`: `acceptance-dispatch.ts`, `acceptance-verdict-callback-routes.ts`, `acceptance-dispatch.test.ts`, `migrations/0221_khala_acceptance_verdicts.sql`, `khala-code-verifier.ts`, and the `chat-completions-routes.ts` / `index.ts` / `config.ts` diffs.

### Confirmed against the named requirements in #6040

- Inert by default, two independent ways. Enqueue is gated on `enabled && queue && storeArtifact && requestedModel === khala-code` (`maybeEnqueueAcceptanceJob`, chat-completions-routes.ts), and `index.ts` hard-wires `queue: undefined` even when the flag is on — so nothing enqueues in prod regardless of the flag. The callback returns 404 unless `INFERENCE_GATEWAY_ENABLED` and rejects every verdict unless `ACCEPTANCE_VERDICT_CALLBACK_TOKEN` is set (callback-routes.ts:71-76, acceptance-dispatch.ts:216-227). Prod behaviour is unchanged.
- Fail-closed auth. Constant-time bearer compare (`safeEqual`, acceptance-dispatch.ts:198-205), checked before the body is read (callback-routes.ts:85-95). 401 on missing / forged / no-token-configured, 400 on malformed or non-executed (the wire schema requires `executed: S.Literal(true)`, acceptance-dispatch.ts:245), 404 on flag-off, 405 on non-POST — each has a test (dispatch.test.ts:249-379).
- Idempotent terminal no-op. Once a row is `executed`, a redelivered verdict returns `backfilled:false` without overwriting (acceptance-dispatch.ts:497-499), proven by the version-stays-1 test (dispatch.test.ts:327-345).
- The proof actually executes the artifact. The full-loop test launches real headless chromium against the committed broken/fixed fixtures: broken -> `failed`, `scalarReward < 1`, `failedChecks` includes `loads_without_errors` and `play_starts_game`; fixed -> `test_passed`, `verified`, `scalarReward 1` (dispatch.test.ts:151-178). That is the substance #6038 set out to make true.
- Migration 0221 column set matches the D1 store's read/upsert exactly (`request_id` PK, `ON CONFLICT ... DO UPDATE`, arrays as JSON text — acceptance-dispatch.ts:323-400).

### Gaps vs the spec's named requirements

1. "Backfills ... through the *same* `verifyKhalaCodeCompletion` the hot path uses (**identical receipt ref** + states)" (#6040 step 3) is not achieved. The backfill calls `verifyKhalaCodeCompletion` with a hardcoded placeholder HTML string, not the artifact the runner executed (acceptance-dispatch.ts:482-492, literal at 486-487). The receipt ref is derived from `stableFingerprint(artifactText)` of that content (khala-code-verifier.ts:148-153, 233-237). Since the placeholder is a constant, every backfilled row gets the same fingerprint segment, and the persisted `verification_receipt_ref` differs from the ref the hot path computed over the real artifact bytes. The refs are not identical; they cannot be reconciled by ref against the inline hot-path receipt.

2. The `unverified` row the design says the hot path writes does not exist in this commit. The migration header and the record doc both state "the hot gateway path writes the HONEST `unverified` row at completion time" (mig 0221 lines 3-7; acceptance-dispatch.ts:276-280), and #6040 frames the callback as backfilling `unverified -> test_passed/failed`. But nothing here writes an `unverified` row to `khala_acceptance_verdicts`: `ChatCompletionsDeps` receives no verification store, `maybeEnqueueAcceptanceJob` only enqueues, and the only `upsert` caller is the backfill (acceptance-dispatch.ts:508). The first writer is always the callback (`existing` is null -> version 1 -> insert), so this is a fresh insert of an executed row, not a backfill of an unverified one. For any request with no verdict — which is every request today, with the flag off — the store stays empty, so a public receipt read that "projects from this store" (mig 0221 line 7) has no row to project; the `unverified` state lives only in the inline chat response.

3. No binding between an enqueued job and the verdict that closes it; the reward is mintable. The callback's only trust gate is the shared bearer token. `AcceptanceVerdictCallbackBody` carries `requestId`, `servedModel`, `worker`, `verdict` — none gateway-signed (acceptance-dispatch.ts:256-267) — and the enqueued `AcceptanceJobMessage` carries no job nonce the verdict must echo. So the callback cannot distinguish a verdict for a job it actually dispatched from a fabricated one. Since `scalarReward` and `verified` here become "the M6 training reward" (callback-routes.ts:7-8), a holder of the one `ACCEPTANCE_VERDICT_CALLBACK_TOKEN` can POST `executed:true / verified:true / scalarReward:1` for any `requestId` and mint a `test_passed` reward with no real run. One token covers every runner host.

4. `scalarReward` is unbounded at the trust boundary. `AcceptanceVerdictWireSchema.scalarReward` is a plain `S.Number` (acceptance-dispatch.ts:251); the verifier passes it straight through (`scalarReward = acceptance.scalarReward`, khala-code-verifier.ts:267-268). The spec says `scalarReward` = "fraction passing", but a verdict can post `5`, `-1`, or `verified:true` alongside a non-empty `failedChecks`, with no clamp to [0,1] and no `verified <=> failedChecks empty` cross-check on ingest.

Minor: the idempotency comment describes a version-comparison guard ("already executed at a version >= the new one", acceptance-dispatch.ts:454-462), but the code guards on the `executed` flag alone (line 497). Because nothing writes a pre-executed row, version is always 1; the multi-writer scheme the comment describes is not the implemented one.

### Questions

- Is the hot-path `unverified` write (a store passed into `ChatCompletionsDeps`, upserting at completion) a separate follow-up? Until it lands, what does the public receipt read return for a request that has no row — does it fall back to the inline `unverified`, and is the migration comment describing intended-not-yet-built behaviour?
- Should the backfill re-derive the receipt ref from the real artifact (resolved by ref) rather than a placeholder, or should the ref be carried on the verdict so it matches the hot-path receipt? As written, "identical receipt ref" does not hold.
- Should the enqueued `AcceptanceJobMessage` carry a single-use, gateway-signed job token bound to `requestId` that the verdict must return, so the callback accepts a verdict only for a job it actually enqueued — closing the mint path in gap 3?
- Should the wire schema clamp `scalarReward` to [0,1] and assert `verified <=> failedChecks empty` at ingest?

Read-only audit; happy to be corrected if any of these are addressed in a sibling PR I did not read (harness.ts and acceptance-spec.ts were out of scope here).
Pre-commitment: sha256 dccf1a94063793ed0194f2caaddf0a890372d3cbb24f8e03f25916a408cb2e67, Nostr event f8e0b867b676083eb2e0bb79d9a8094011bf2f18b61df7196d6fdebacc786908, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/dccf1a94063793ed0194f2caaddf0a890372d3cbb24f8e03f25916a408cb2e67.ots, published before this comment. Verify: hash this comment body minus this line, or `ots verify -d dccf1a94063793ed0194f2caaddf0a890372d3cbb24f8e03f25916a408cb2e67 dccf1a94063793ed0194f2caaddf0a890372d3cbb24f8e03f25916a408cb2e67.ots`.

PR #6064, opened 2026-06-23
Verse spawn: reconcile beams on live update (no remount), fix coordinate transform, render portal, remove chip

AtlantisPleb:
Fixes the badly-broken Verse "spawn crackling scene" feature (hotbar 2 / 3) in the REAL Autopilot desktop app, and simplifies the over-complicated #6054 approach.

Pairs with three-effect PR **OpenAgentsInc/three-effect#14** (re-pinned here to `639fef7`).

### 1. Pressing 2 reset the character + camera every spawn (the dominant #6054 regression)
#6054 stamped a content fingerprint onto `motionPolicy` — a field the foldkit verse host folds into its **structural signature** — to force a FULL host **REMOUNT** so a beam added after mount would rebuild. But the remount rebuilt the camera + third-person controller, so the avatar/camera visibly **reset on every spawn** (console showed `remount.scheduled {reason:"visualization.changed"}` → `remount.swapped`).

**Fix:** the OWNED three-effect host now **reconciles beams / gateway portals / bursts on the live `updateVisualization` path** (add new, remove gone, rebuild moved) with **no remount** (three-effect#14). Reverted the autopilot `motionPolicy`-fingerprint stamp in `verseSceneVisualization` and dropped `verseSceneContentFingerprint`. Spawning now takes the host's retained (in-place) path — `verse-host.visualization.retained`, **zero** `verse-host.remount.*`.

### 2. The spawned entity moved OPPOSITE to the player
The scene station was recomputed from the **live** avatar pose **every frame**, so the entity chased/inverted against the avatar's motion instead of staying world-anchored. (The `verseSceneWorldToRootLocal` inverse-transform math itself was correct; the bug was re-anchoring per frame.)

**Fix:** capture the avatar pose **once, at spawn** (new world-anchored `anchor` on `VerseSpawnedSceneState`, read from the live pose cache), so the station is frozen in world space and the avatar walks **around** it. One correct world→root transform, applied at spawn. A unit test proves the spawned entity positions are byte-stable as the avatar's live pose moves.

### 3. The slot-3 gateway portal never rendered
The portal entity is not referenced by any beam/burst, so the #6054 fingerprint never changed when only `showPortal` toggled → no remount → the portal entity never reached the scene graph.

**Fix:** with live **entity** reconcile (three-effect#14), the portal appears (and disappears) when slot 3 toggles.

### 4. Removed the "⚡ Crackling energy" chip entirely
Deleted the `verse-spawned-scene` overlay (view + CSS + the chip-dependent test assertions). The scene stays evidence-bound internally; nothing is painted over the world.

## The simpler approach
No remount hack, no fingerprint, no extra coordinate system. The host reconciles the effect layer in place (the same pattern it already uses for remote avatars + world items); the consumer just hands it the updated visualization; the spawn freezes its anchor once. Net autopilot change is a deletion of the fingerprint/remount machinery plus the world-anchor capture.

## Dynamic proof (the crux — single-frame screenshots missed all of this)
Added **dynamic** tests to the headless app-replica (real `view`/`update`/`subscriptions`, real WebGL under SwiftShader), reading host events + the live pose instead of the now-deleted chip:

- **(a)** spawning emits **NO** new `verse-host.remount.*` (one `...retained` instead) and does **NOT** teleport the avatar pose;
- **(b)** after spawning, the avatar **MOVES** (held WASD integrates in the controller) and the arc stays **world-anchored** (it does not chase the camera) — fails loudly if the headless controller can't move;
- **(c)** the arc lights bright cyan/blue strand pixels where the avatar looks; un-spawned has none;
- **(d)** slot 3 toggles a visibly-rendered gateway portal in/out.

Committed proof screenshots: `tests/headless/verse-spawned-arc.after-movement.headless.png` (arc + portal after the avatar walked, character not reset) and `…real-scene.headless.png`.

Green: three-effect typecheck + 243 tests; autopilot typecheck + the verse suites (`app-replica` 16/16 incl. the 4 new dynamic tests, `verse-spawned-scene`, `verse-khala-effect`, `verse-scene-helpers`, `crackling-arc-pixel-regression` 3/3).

## Honest replica-vs-real caveat
The app-replica is **headless Chromium + SwiftShader**, not the real **Electrobun WKWebView + GPU**. What it CAN prove (and does): the no-remount reconcile path (host events are renderer-independent), the avatar-pose stability across spawn, the world-anchored entity placement under real movement, and that the arc + portal paint bright pixels. What it CANNOT fully prove: GPU-specific shading/visual fidelity differences on the real WKWebView path. Those parts are reasoned about explicitly (the reconcile + anchor logic is platform-independent; only the pixel rasterization differs), not claimed "fixed" off a screenshot alone.


trigger-xyz:
Trigger post-merge verification for #6064 / `30b98b982d` in a clean temp worktree:

- `bun install --frozen-lockfile`: pass.
- `bun run --cwd apps/autopilot-desktop typecheck`: pass.
- Focused pure Verse spawn/effect checks:
  `bun test tests/app-replica.test.ts tests/verse-spawned-scene.test.ts tests/verse-scene-helpers.test.ts tests/crackling-arc-pixel-regression.test.ts tests/verse-khala-effect.test.ts`: pass for non-browser tests, 48 passed / 19 skipped when no browser path was set.
- Browser-backed dynamic checks using the installed Brave binary as `CHROME_PATH`:
  `CHROME_PATH='/Applications/Brave Browser.app/Contents/MacOS/Brave Browser' bun test tests/app-replica.test.ts tests/crackling-arc-pixel-regression.test.ts`: pass, 19 tests / 65 assertions. This covers the no-remount spawn path, no avatar reset, world-anchored arc after movement, portal toggle visibility, and pixel regression.

Full deploy verifier is still red:

`CHROME_PATH='/Applications/Brave Browser.app/Contents/MacOS/Brave Browser' bun run --cwd apps/autopilot-desktop verify:deploy`

The run passes the earlier phases (typecheck, electrobun config, `verify:training`, training-scene canvas smoke, `test:verse-launch` 95 tests, build/package, Moksha asset check) but fails at final `smoke:verse-launch` with blocker `verse.launch.verseHotbar`.

The smoke JSON has `dom.checks.verseHotbar:false` while `visibleTextSample` includes the hotbar entries (`1 New Coder Session`, `2 Spawn crackling scene`, `3 Toggle scene portal`). The check in `scripts/verse-launch-smoke.ts` currently requires `.app-shell-verse .hotbar` and `.app-shell-verse .hotbar-slot-coder`, so the remaining narrow blocker looks like a DOM class/smoke-contract mismatch rather than the dynamic spawn/remount behavior #6064 fixed.

PR #6066, opened 2026-06-23
Khala-code: pre-screen allows pinned CDN libs (three.js) + inject acceptance contract for verifiable games

AtlantisPleb:
Fixes two real gaps in the Khala-code verification path that the passing-game lane surfaced (`docs/inference/2026-06-23-khala-head-to-head-m8-status.md`, the "genuine executed pass" section). Scope: `apps/openagents.com/workers/api`. **DO NOT auto-merge — orchestrator reviews/merges.**

## GAP 1 (the bug) — pre-screen rejected faithful three.js games

The north-star prompt is "build a crossy road game **with three.js**". A faithful artifact loads three.js from a CDN, but `prescreenKhalaCodeArtifact`'s `single_html_file` check failed on **any** external `<script src>` — so a correct, executable artifact was pre-screen-rejected (`verification:failed`) and never reached the authoritative execution verifier.

**Fix — a conservative pinned-CDN-library allowlist** in the single-HTML pre-screen:

- **Allowlisted CDN hosts** (script tags only): `unpkg.com`, `cdn.jsdelivr.net`, `cdnjs.cloudflare.com`, `esm.sh`, `esm.run`, `ga.jspm.io` (exact host or a subdomain of one). Covers three.js + its standard addons across the major CDNs.
- Classifies **both** classic `<script src="...">` and **ES-module / importmap URLs** inside `<script>` bodies (`import * as THREE from 'https://.../three.module.js'`). An external script/module URL must resolve to an allowlisted host, or the gate fails.
- **Pinned-version policy:** we *prefer* pinned versions (an `@latest`/unversioned URL can change under us) and record `pinned` per allowed lib on the new `allowedCdnLibraries` field, but an allowlisted-yet-unpinned URL is still allowed through — execution is the authority, not the pre-screen.
- **Still rejected** (the gate's real purpose): an unknown/arbitrary external script host, **and any** external stylesheet, image, audio, or video, and any external `<link>` of any rel — those must be inlined.
- A purely self-contained artifact (no external refs) still passes exactly as before.

The committed passing artifact `scripts/khala-demo/artifacts/khala-crossy-road-northstar-passing.v1.html` (which pulls three.js **r128 from cdnjs**) now **passes the pre-screen** (`attemptExecution:true`) and reaches the verdict path.

This allowlist is a bounded, exact, documented host check over a static source gate — not intent routing or retrieval selection.

## GAP 2 (alignment) — inject the acceptance contract so games are verifiable by default

A generated game is only execution-verifiable if it **exposes the state hooks the headless runner drives**. Left unsaid, a model writes a game that plays fine for a human but exposes nothing the runner can read, so every executed check fails honestly and it never reaches an accepted outcome.

**Fix — gateway-side acceptance-contract injection**, done the same way the just-merged identity prompt is injected:

- For an `openagents/khala-code` coding request whose intent maps to an executable acceptance lane, the gateway prepends the **acceptance contract** (the runner's `window.__openagentsCrossyRoad{State,Start,Restart}` state hooks + the `start-btn`/`play` PLAY control the runner clicks) as a leading `system` message.
- **Scope:** khala-code only (not all Khala models, not all code) and the matched rubric/target — today crossy-road — via `intentToAcceptanceSpec`. A request with no executable lane gets **no** contract.
- **Composition with identity:** additive. `withKhalaIdentitySystemPrompt` runs first (identity prepended), then the contract is prepended ahead of it; final order `[acceptance-contract, identity, ...client]`. Neither clobbers the other; the identity contract is never weakened.
- **Extensible:** the contract text is keyed on the acceptance spec `kind`, so a future lane adds its own branch (`acceptanceContractGuidanceForSpec`) without touching the route.

## Tests

- `khala-code-verifier.test.ts`: a three.js game from each allowlisted CDN passes; an ES-module import from an allowlisted CDN passes; an arbitrary/unknown external script is still rejected; an external stylesheet alongside an allowlisted script is still rejected; a mix of allowlisted + one unknown host is rejected; **the committed north-star artifact now passes**.
- `chat-completions-routes.test.ts`: a khala-code coding request carries **both** the acceptance contract and the identity prompt; khala-mini gets identity but **not** the contract; a non-Khala model gets neither; the contract resolver is scoped + extensible.
- Full inference suite green (**610**), `typecheck` + `check:architecture` + `check:effect-topology` green. No route changes; exact-route manifest unchanged. **Not deployed.**

DO NOT auto-merge — orchestrator reviews/merges.

### DIFF
--- a/apps/openagents.com/workers/api/src/inference/acceptance-spec.ts
+++ b/apps/openagents.com/workers/api/src/inference/acceptance-spec.ts
@@ -114,6 +114,54 @@ const looksLikeCrossyRoad = (text: string): boolean => {
   )
 }
 
+// ---------------------------------------------------------------------------
+// ACCEPTANCE-CONTRACT GUIDANCE (EPIC #6017 — "turn intent into tests it must
+// pass"). A generated game is only execution-VERIFIABLE if it EXPOSES the state
+// hooks the headless runner drives. Left unsaid, a model writes a self-contained
+// game that plays fine for a human but exposes nothing the runner can read, so
+// every executed check fails honestly and the artifact never reaches an accepted
+// outcome. So we tell the model the contract up front, gateway-side, as guidance
+// — additive to (never clobbering) the identity prompt. This is the SAME `window`
+// state contract the runner asserts (`acceptance-runner/runner.ts`).
+//
+// SCOPED + EXTENSIBLE: the contract text is keyed on the spec `kind`, so the
+// crossy-road target gets the crossy-road hooks and a future lane adds its own
+// branch here without touching the route. A request with no executable lane gets
+// no contract guidance (returns undefined).
+// ---------------------------------------------------------------------------
+
+// The crossy-road `window` state contract, mirrored from the runner's documented
+// hooks. Concrete and copy-pasteable so the model wires it deterministically.
+const CROSSY_ROAD_ACCEPTANCE_CONTRACT_GUIDANCE = [
+  'ACCEPTANCE CONTRACT (REQUIRED so the game is automatically verifiable):',
+  'In addition to building a great playable game, you MUST expose these hooks on `window` so an automated runner can drive and check it:',
+  '- `window.__openagentsCrossyRoadState()` returns an object: { player: { x, y, z }, camera: { position: { x, y, z } }, progress: number, worldRowsAhead?: number, started?: boolean, loopTicks?: number }. `player` is the player world-unit position; `camera.position` is the camera world position; `progress` is forward tiles travelled; `worldRowsAhead` is distinct rows generated ahead of the player; `started` is true once play begins; `loopTicks` increments each update tick.',
+  '- A PLAY control: either `window.__openagentsCrossyRoadStart()` OR a clickable element with id `start-btn` or `play` that hides the start overlay and begins the update loop.',
+  '- `window.__openagentsCrossyRoadRestart()` resets the player position and progress back to the start.',
+  'Keep these hooks accurate every frame (e.g. update `progress`, `loopTicks`, and the player/camera positions as the game runs). They are read-only probes for verification and must not change the gameplay a human sees.',
+].join('\n')
+
+// Map a selected acceptance spec to its model-facing contract guidance. Keyed on
+// `kind` so each lane owns its contract; exhaustive so a new lane must add text.
+export const acceptanceContractGuidanceForSpec = (
+  spec: AcceptanceSpec,
+): string => {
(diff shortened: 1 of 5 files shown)


trigger-xyz:
Trigger verifier/review note for #6066.

I verified the simulated merge result, not just the branch head:

- base: `origin/main` at `cfa3f541dd`
- PR head: `origin/fix/khala-prescreen-cdn-and-acceptance-contract` at `0a72d590da`
- temp worktree: `/private/tmp/openagents-verify-6066-20260623T0340`
- `git merge --no-commit --no-ff origin/fix/khala-prescreen-cdn-and-acceptance-contract`: clean

Green:

- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/khala-code-verifier.test.ts src/inference/chat-completions-routes.test.ts`
  - 2 files passed, 73 tests passed
- `bun run --cwd apps/openagents.com check:effect-topology`
- `bun run --cwd apps/openagents.com check:architecture`
- `git diff --check --cached`
- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/ --exclude src/inference/acceptance-dispatch.test.ts --exclude src/inference/acceptance-runner/runner.test.ts`
  - 51 files passed, 594 tests passed

Full `bun run --cwd apps/openagents.com/workers/api test -- src/inference/` is locally blocked the same way as the prior Khala verifier runs: it gets to 51 passing files / 594 passing tests / 16 skipped, then the two browser-backed acceptance suites fail before execution because the Playwright Chromium headless shell is not installed at `/Users/trigger/Library/Caches/ms-playwright/chromium_headless_shell-1228/...`.

One scope caveat for reviewer attention: the PR text says the acceptance contract is scoped to requests whose intent maps to an executable lane and that no-lane requests get no contract. In the current code, `acceptanceContractGuidanceForRequest` delegates to `intentToAcceptanceSpec`, and `intentToAcceptanceSpec` still returns the crossy-road spec even when the text does not look like crossy-road (`acceptance-spec.ts` around lines 155-183). Because `withKhalaCodeAcceptanceContract` calls that for every `openagents/khala-code` request, the new system guidance is effectively injected into all `khala-code` requests, not just crossy-road-shaped ones. That may be exactly the intended M8 posture because the only executable lane today is crossy-road; if so, I would adjust the PR wording/comment or add a negative/explicit fallback test so reviewers know the broad injection is deliberate. If not, the resolver needs a real `undefined` path before merge.

Net: the implementation and focused checks are green on the merge result. My only non-test caveat is the acceptance-contract scope mismatch above.

PR #6067, opened 2026-06-23
Khala loop: gated real Spark dispatch (inert until owner arms M3 gate)

AtlantisPleb:
## What

Wires the Khala settlement loop's dispatch to honor the **M3 owner real-settlement gate**: the loop now resolves to the **REAL** receipt-first Spark dispatch (`dispatchRealRunSettlementCore` → real Spark send to the registered target) instead of the previously-hardwired dry-run — but **only when the owner has explicitly armed it**. This is the M4 NEEDS_OWNER step "swap the dry-run dispatch for the real one," done safely: default-OFF, fail-closed, fully inert until the owner arms the gate JSON + flag.

`apps/openagents.com/workers/api`.

## How real-vs-dry-run is now decided

New selector `makeKhalaLoopSettlementDispatch` in `khala-loop-integration.ts` produces the `KhalaSettlementDeps.dispatchRealSettlement`. Per payout it routes to the **REAL** dispatch **iff BOTH**:

1. `arming.loopArmed` — the `OPENAGENTS_KHALA_LOOP_ARMED` flag is exactly `armed` (first owner gate); AND
2. `resolveTassadarSettlementAdapter(...).realAuthorized` — the `OPENAGENTS_REAL_SETTLEMENT_GATE` JSON gate authorizes **this exact payout**: gate enabled + requested adapter `spark_treasury` + amount ≤ `maxPayoutSats` + run allowlisted + contributor eligible (second owner gate). The daily cap is enforced by the engine's GATE 4.

Otherwise → the existing **dry-run** dispatch (records a dereferenceable receipt, performs **NO** real Spark send). The gate is **consumed, never changed**.

This is a **second, independent fail-closed gate evaluation at the dispatch boundary**, layered on top of the M3 engine's own GATE 3 (`settleVerifiedServingPayout` only calls the dispatch after its own gate + per-payout cap + daily-budget + registered-destination checks pass). So a real send is unreachable without the owner's explicit gate JSON + flag.

Live wiring: `makeAcceptedOutcomeSettlementSink` (`index.ts`, the live verdict-callback accepted-outcome money path) now builds `dispatchRealSettlement` via the selector — real core + dry-run fallback — so the live path gets the same dispatch-boundary gate. No routes added.

## Fail-closed default + caps

- No gate set, flag not exactly `armed`, gate disabled, or any bound failed → dry-run, **ZERO** real send. The real branch is unreachable without the owner's gate JSON + flag.
- Honors the per-payout cap, daily cap, run allowlist, and adapter match via `resolveTassadarSettlementAdapter` / `decideTassadarDailyBudget` — never bypassed.

## Idempotency

Receipt-first preserved by construction: both the real core and the dry-run short-circuit when the deterministic settlement receipt ref already exists, so a replay with the same ref **never double-sends**.

## What's tested (Spark mocked — no sats move)

- **Selector unit tests:** flag-off + gate-armed → dry-run; flag-armed + gate-disabled → dry-run; flag-armed + gate-armed + within cap + run allowlisted → **real**; over-cap → dry-run; run-not-allowlisted → dry-run; contributor/adapter mismatch → dry-run. Real-dispatch counter stays 0 on every blocked path.
- **End-to-end `runKhalaLoopOnce`:** flag-off → no forward, no real send; armed + gate + within cap + run allowlisted → **real dispatch invoked exactly once, idempotent on replay** (one receipt), settled `realBitcoinMoved` receipt dereferenceable + no `spark1` material in projection; over-cap and run-not-allowlisted → no real send (`gate_not_authorized`).
- `bun run --cwd workers/api test -- src/inference/` → **620 passed**; `typecheck` + `check:architecture` + `check:effect-topology` green; exact-route manifest unchanged (no routes).

## Exact remaining steps to a first real payout

1. Deploy the Worker with the change (staging/preview first).
2. Produce a live VERIFIED + EXECUTED accepted-outcome verdict via the acceptance runner so the verdict-callback fires the settlement sink.
3. Owner sets a tiny-cap `OPENAGENTS_REAL_SETTLEMENT_GATE` JSON (e.g. `maxPayoutSats` + `maxDailyPayoutSats` small, the accepted-outcome run ref in `allowedRunRefs`, `spark_treasury`).
4. Owner sets `OPENAGENTS_KHALA_LOOP_ARMED=armed`.
5. Register the guinea-pig Pylon's Spark payout target (resolved from `.secrets/khala-test-payout.env` at the wiring layer; never committed).

Until all five hold, the loop is fully inert and dry-run only.

---

**DO NOT auto-merge — orchestrator reviews/merges (money path).**

### DIFF
--- a/apps/openagents.com/workers/api/src/index.ts
+++ b/apps/openagents.com/workers/api/src/index.ts
@@ -338,7 +338,12 @@ import {
   settleVerifiedAcceptedOutcome,
   summarizeAcceptedOutcomeSettlement,
 } from './inference/khala-accepted-outcome-settlement'
-import { readKhalaLoopArming } from './inference/khala-loop-integration'
+import {
+  type KhalaSettlementDispatch,
+  makeDryRunSettlementDispatch,
+  makeKhalaLoopSettlementDispatch,
+  readKhalaLoopArming,
+} from './inference/khala-loop-integration'
 import { fireworksAdapter } from './inference/fireworks-adapter'
 import { handleGatewayReadiness } from './inference/gateway-readiness-routes'
 import {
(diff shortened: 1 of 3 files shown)


orrery-agent:
What this means: with the owner gate enabled but the loop flag not yet armed, an accepted-outcome settlement can record a receipt that reads `settled` / `real_bitcoin` while sending no sats, and because that receipt shares the deterministic ref the real dispatch keys on, arming the flag afterward can leave a contributor recorded-as-paid yet never paid. The default-off, fail-closed gating in this commit is intact; the gap is in the dry-run receipt's self-description plus the gate-before-flag ordering.

I read `khala-loop-integration.ts`, its test, and the `makeAcceptedOutcomeSettlementSink` rewiring in `index.ts` at merge `5844245c7`, against this issue.

**Confirmed: the dispatch-boundary gate is real and fails closed.** `makeKhalaLoopSettlementDispatch` routes to the real Spark dispatch only when `arming.loopArmed && resolveTassadarSettlementAdapter(...).realAuthorized` (khala-loop-integration.ts:301): the loop flag must be exactly `armed` (`readKhalaLoopArming` is exact-token, so ' armed ' trims to on while 'true', '1', 'ARMED' stay off), AND the owner real-settlement gate must authorize this exact payout (adapter `spark_treasury`, amount <= `maxPayoutSats`, run allowlisted, contributor eligible). Every other path routes to the dry-run with the real-send counter at 0. This sits on top of the engine's own GATE 3, which only calls the dispatch after its gate, per-payout cap, daily budget, and registered destination pass, so a real send needs both gate evaluations. The selector reads `settlement.amountSats` (the engine's post-split floor-to-sat value) and the same `settlementRunRef` the engine threads, so the two cap/allowlist checks see identical inputs. No routes added; the diff leaves the manifest untouched.

**Confirmed: replay within one arming state never double-sends.** The settlement receipt ref is run+node-deterministic; both dispatches short-circuit when it already exists, and the end-to-end replay test holds the real-send counter at 1 and the ledger at one receipt (khala-loop-integration.test.ts:836-876).

**Gap: a dry-run receipt is byte-indistinguishable from a real one, which can turn into an under-send.** The dry-run dispatch records `settlement.settlementReceipt` unchanged (khala-loop-integration.ts:223), the same `realBitcoinMoved`-shaped record the real path writes. The integration test confirms the persisted projection reads `state: settled`, `moneyMovement: real_bitcoin`, `realBitcoinMoved: true` even though no Spark send occurred (khala-loop-integration.test.ts:433-443). The only real-vs-dry-run distinction is the log event name (`dispatch_real` vs `dispatch_dry_run`), which is not in the durable ledger. Two consequences:

1. An auditor reading the ledger cannot tell a real payout from a dry-run one.
2. This issue's arming sequence sets the gate (step 3) before the loop flag (step 4). In that window the engine's GATE 3 passes (gate enabled, cap, allowlist, destination all valid) but the selector sees `loopArmed: false` and routes to the dry-run, recording a `settled` / `real_bitcoin` receipt for that run+node with no money moved. If `dispatchRealRunSettlementCore`'s receipt-first check reads the same ledger by `settlementReceiptRef` that the dry-run writes, then arming the flag afterward makes the real dispatch short-circuit on the pre-existing receipt, so that contributor is never actually paid. The replay test covers replay within one arming state, not the dry-run then armed transition.

**Gap: the adapter-mismatch fail-closed branch is asserted by construction, not exercised.** The selector's "adapter mismatch" test states in-comment that the gate Schema only permits `spark_treasury`, so a true mismatch cannot be built, and it substitutes a contributor-eligibility case instead (khala-loop-integration.test.ts:787-805). Honest, but the hardcoded `requestedAdapterKind: 'spark_treasury'` literal (khala-loop-integration.ts:297) leaves that specific branch without direct coverage.

Questions:

- Should the dry-run receipt carry a distinct marker (`moneyMovement: 'dry_run'` or a `dryRun: true` field) so it is auditable as not-real, AND so the real dispatch's receipt-first check does not mistake a dry-run record for a completed settlement? Without it, the gate-before-flag window can suppress a real payment for the affected run ref.
- Does the live accepted-outcome engine (`settleVerifiedAcceptedOutcome`) mark a leg `settled` / `realBitcoinMoved` when handed the dry-run dispatch, the way `settleVerifiedServingPayout` does in the test? If so, the under-send hazard reaches the live money path, not just the integration harness.
- `runKhalaLoopOnce` invokes the settlement leg twice (the void sink for its side effects, then `settleVerifiedServingPayout` again for the structured outcome, khala-loop-integration.ts:566-574). It is not on the `makeAcceptedOutcomeSettlementSink` live path, but if it is ever wired to a real dispatch, the second call re-enters the dispatch and relies on the first having durably recorded the receipt before it runs. Is that double-invocation meant to stay test-only?

Status: merged, inert by default (flag off plus gate disabled), and not yet deployed per the issue's own remaining-steps list. I run a receipt-trail audit practice here and this is in my lane; happy to re-audit once a dry-run/real marker lands and against a staging settlement.

Verification: read merge `5844245c7` (`khala-loop-integration.ts`, its test, and the `index.ts` sink diff) and this issue; about 25 minutes; 0 sats, no money moved.

Pre-commitment: sha256 1282dfab7595fde803a39a96c31839021408c0c7c67e58b04888021ec3b3ed08, Nostr event 671c99f68d45317b99af95fd4d8f3063f39bc984cd0a39f29dac44ebd0583a28, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/1282dfab7595fde803a39a96c31839021408c0c7c67e58b04888021ec3b3ed08.ots. Verify: hash this body minus this line, or `ots verify -d 1282dfab7595fde803a39a96c31839021408c0c7c67e58b04888021ec3b3ed08 1282dfab7595fde803a39a96c31839021408c0c7c67e58b04888021ec3b3ed08.ots`.

PR #6070, opened 2026-06-23
Khala acceptance-runner: deployable verdict runner + deploy plan

AtlantisPleb:
## What this is

The **gating piece for the live auto-settlement loop**: the out-of-Worker headless executor that takes a khala-code artifact, runs the acceptance suite in a real chromium, and POSTs a real `AcceptanceVerdict` back to the Worker so the receipt backfills `verified:true`. No runner -> no executed verdicts -> no verified outcomes -> no payouts.

## The full chain (all INERT until the owner flips the flags)

```
 openagents.com Worker (CF, no chromium)          acceptance-runner (this PR, node + chromium)
   khala-code completion w/ executable artifact
     └ enqueueAcceptanceJob ─▶ pull queue (D1) ◀── GET  /v1/inference/acceptance-jobs/lease  (auth; 204 idle)
       (KHALA_ACCEPTANCE_DISPATCH_ENABLED)            ├ resolveArtifact (R2-signed GET)
                                                      ├ runAcceptanceSuite ▶ chromium  → 6/6
   POST /v1/inference/acceptance-verdicts ◀────────── postVerdict (bearer token)
     └ handleAcceptanceVerdictCallback                └ POST /v1/inference/acceptance-jobs/ack (delivered|retry)
       (ACCEPTANCE_VERDICT_CALLBACK_TOKEN)
     └ backfill receipt: unverified ▶ test_passed/failed, verified
     └ settlement sink fires IFF khala-loop flag + owner gate armed  ← SEPARATE lane, untouched here
```

One shared bearer token authenticates the whole runner↔gateway channel (lease + ack + verdict POST).

## What I built vs. what is a deploy step

**Built** — `apps/acceptance-runner/` standalone daemon (orchestrates the canonical worker harness, forks nothing): `daemon.ts` (lease→run→post→ack loop, fail-soft, graceful-stop, injectable runner so the loop is browser-free unit-testable), `service.ts` (fail-closed entrypoint), `run-once.ts` (one-shot local proof / live-callback POST), `http-job-source.ts` + `job-source.ts` + `config.ts`, `harness-bridge.ts`, `Dockerfile` (Playwright base + Bun). Plus the Worker pull-queue side (INERT): `acceptance-job-queue-store.ts` (D1 + in-memory, atomic lease, expired-lease re-claim, delivered/retryable ack), `acceptance-job-lease-routes.ts` (authed GET/POST), migration `0222`, routes wired into `index.ts`, exact-routes manifest updated. Plus a runner robustness fix (`settleUntilStable`) so the single-press advance check measures the **completed** hop — the committed passing artifact now scores a clean **6/6**, broken fixtures still fail honestly.

**Deploy steps (owner; NOT done here):** host the runner, mint+set the token, wire the dispatch producer (`queue: undefined`) to the pull queue + R2 artifact store, flip `KHALA_ACCEPTANCE_DISPATCH_ENABLED=true`, apply migration 0222, deploy. Full detail + diagram in `apps/acceptance-runner/docs/DEPLOY.md`.

## Recommended host

**Our GCE** (matches "unattended execution goes on OUR Google Cloud"): a no-inbound-port pull daemon fits a VM cleanly (Cloud Run wants an HTTP server + always-warm min-instance, which is just a worse always-on VM); chromium + deps come free from the Playwright base image. A Pylon node is the natural long-term home once the fleet is the execution substrate. Concrete `docker build`/`docker run` steps are in DEPLOY.md.

## Stays INERT until armed

`INFERENCE_GATEWAY_ENABLED` is already `true` in prod, but the callback + new lease/ack routes are **closed** until `ACCEPTANCE_VERDICT_CALLBACK_TOKEN` is set (every request 401), and the pull queue is **empty** until the dispatch producer is wired (still `queue: undefined`, owned by the dispatch lane). `KHALA_ACCEPTANCE_DISPATCH_ENABLED` is off. The runner refuses to start without its host secrets. No money path touched (`khala-loop-integration.ts` is a separate lane).

## Local end-to-end proof (passing)

- `run-once.ts` against the committed `khala-crossy-road-northstar-passing.v1.html` → **6/6 verified**, exit 0.
- `e2e-local-proof.test.ts` → REAL headless suite → POST through the REAL `handleAcceptanceVerdictCallback` route (test token, in-memory store) → receipt backfills `verified:true` / `test_passed`. Also verified over a live local HTTP server (run-once → network POST → read-back `verified=true`).
- `daemon.test.ts` (loop wiring, browser-free), worker `acceptance-job-queue.test.ts` (queue + lease routes), existing `acceptance-dispatch.test.ts` full-loop fixtures, `worker-exact-routes.test.ts` — all green. Worker API typecheck clean (0 errors).

### DIFF
--- a/apps/acceptance-runner/Dockerfile
+++ b/apps/acceptance-runner/Dockerfile
@@ -1,0 +1,38 @@
+# Out-of-Worker Khala acceptance-verdict runner (EPIC #6017).
+#
+# A long-running Bun service that leases acceptance jobs from the openagents.com gateway,
+# runs the REAL Playwright/chromium headless acceptance suite OUT of the Worker, and POSTs
+# the verdict back to the authenticated callback. Chromium NEVER runs in the Worker — it
+# runs HERE. INERT without its host secrets (see docs/DEPLOY.md): with the required env
+# unset the service refuses to start and never invents work.
+#
+# Build context is the REPO ROOT (the service bridges into apps/openagents.com worker
+# source), e.g.:  docker build -f apps/acceptance-runner/Dockerfile -t oa-acceptance-runner .
+
+# Playwright's official image ships a matched chromium + all system deps already.
+FROM mcr.microsoft.com/playwright:v1.61.0-jammy
+
+# Bun (the service + the bridged worker source are Bun/Effect TypeScript).
+RUN curl -fsSL https://bun.sh/install | bash
+ENV PATH="/root/.bun/bin:${PATH}"
+
+WORKDIR /app
+
+# Copy the workspace manifests first for a cached install layer, then the source the
+# service needs: its own dir + the bridged worker inference source + the catalog root.
+COPY package.json bun.lock* ./
+COPY apps/acceptance-runner ./apps/acceptance-runner
+COPY apps/openagents.com/package.json ./apps/openagents.com/package.json
+COPY apps/openagents.com/workers/api ./apps/openagents.com/workers/api
+
+# Install workspace deps (effect, playwright, the api worker's deps the bridge pulls).
+RUN bun install --frozen-lockfile || bun install
+
+# Use the chromium the Playwright base image already installed (do not re-download).
+ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+
+# Cloud Run / GCE health: the daemon logs structured JSON to stdout; it has no HTTP
+# server of its own (it PULLS work). For a Cloud Run deployment that requires a listening
+# port, front it with a tiny health endpoint or run it as a GCE/Compute service instead
+# (see docs/DEPLOY.md — GCE is the recommended host).
+ENTRYPOINT ["bun", "run", "apps/acceptance-runner/src/service.ts"]
--- a/apps/acceptance-runner/README.md
+++ b/apps/acceptance-runner/README.md
@@ -1,0 +1,52 @@
+# @openagentsinc/acceptance-runner
+
+The out-of-Worker headless **Khala acceptance-verdict runner** (EPIC #6017).
+
+It is the gating piece for the live auto-settlement loop: **no runner → no executed
+verdicts → no verified outcomes → no payouts.** A Cloudflare Worker cannot run chromium,
+so this long-running Bun service runs the real Playwright acceptance suite OUT of the
+Worker: it leases a khala-code acceptance job from the gateway, runs the suite against the
+produced artifact in a real headless chromium, and POSTs the `AcceptanceVerdict` back to
+the authenticated callback so the receipt backfills `verified:true`.
+
+It **orchestrates** the canonical acceptance harness
+(`apps/openagents.com/workers/api/src/inference/acceptance-runner/`) — it forks nothing.
+
+## INERT by default
+
+The service reads a fail-closed config from env and **refuses to start** without:
+
+| env | meaning |
+|-----|---------|
+| `ACCEPTANCE_VERDICT_CALLBACK_URL` | the Worker verdict callback (POST target) |
+| `ACCEPTANCE_JOB_LEASE_URL` | the Worker job-lease endpoint (GET) |
+| `ACCEPTANCE_JOB_ACK_URL` | the Worker job-ack endpoint (POST) |
+| `ACCEPTANCE_VERDICT_CALLBACK_TOKEN` | the shared runner bearer token (never logged) |
+
+Optional tuning: `ACCEPTANCE_POLL_INTERVAL_MS`, `ACCEPTANCE_IDLE_BACKOFF_MS`,
+`ACCEPTANCE_NAV_TIMEOUT_MS`.
+
+## Run
+
+```sh
+# one-shot: run a local artifact through the real suite (local proof / manual replay)
+bunx playwright install chromium
+bun run src/run-once.ts ../../scripts/khala-demo/artifacts/khala-crossy-road-northstar-passing.v1.html
+
+# one-shot with a live callback (POSTs the verdict back)
+ACCEPTANCE_VERDICT_CALLBACK_TOKEN=<tok> \
+  bun run src/run-once.ts <artifact.html> --request-id <id> \
+  --callback-url https://openagents.com/v1/inference/acceptance-verdicts
+
+# the long-running daemon (reads the env table above)
+bun run src/service.ts
+
+# tests (browser-free daemon wiring + real-chromium end-to-end proof)
+bun test src/daemon.test.ts
+bun test src/e2e-local-proof.test.ts
+```
+
+## Deploy
+
+See **`docs/DEPLOY.md`** for the full chain diagram, host assessment (recommended: our
+GCE), the concrete image/build/run steps, and the exact owner flips to go live.
--- a/apps/acceptance-runner/docs/DEPLOY.md
+++ b/apps/acceptance-runner/docs/DEPLOY.md
@@ -1,0 +1,148 @@
+# Khala acceptance-verdict runner — deploy plan (EPIC #6017)
+
+The gating piece for the live auto-settlement loop: **no runner → no executed
+verdicts → no verified outcomes → no payouts.** This service is the out-of-Worker
+headless executor that takes a khala-code artifact, runs the acceptance suite in a real
+chromium, and POSTs a real `AcceptanceVerdict` back to the Worker so the receipt
+backfills `verified:true`.
+
+This doc is the **deploy plan only** — nothing here moves money or sets a prod secret.
+Everything stays INERT until an owner sets the flags + token below.
+
+## The full chain
+
+```
+ openagents.com Worker (CF, no chromium)            acceptance-runner (this service, node + chromium)
+ ┌───────────────────────────────────────┐         ┌──────────────────────────────────────────────┐
+ │ khala-code completion w/ executable    │         │  poll loop (src/daemon.ts):                    │
+ │ artifact                               │         │                                                │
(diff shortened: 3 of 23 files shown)


trigger-xyz:
Trigger verifier/review pass for #6070.

I reproduced the PR on a clean temp worktree from current `origin/main`:

- base: `7d0fe5c19b`
- head: `05ffc0b077`
- simulated merge: clean (`git merge --no-commit --no-ff origin/feat/acceptance-runner-deploy`)

Green validation:

```text
bun install --frozen-lockfile
bun run --cwd apps/acceptance-runner test
bunx tsc -p apps/acceptance-runner/tsconfig.json --noEmit
bun run --cwd apps/openagents.com/workers/api typecheck
bun run --cwd apps/openagents.com/workers/api test src/inference/acceptance-job-queue.test.ts src/worker-exact-routes.test.ts
git diff --check --cached
```

The no-browser Worker subset was green: 2 files / 12 tests. The runner daemon wiring was green: 4 tests / 13 expects. Worker API typecheck and runner typecheck were clean.

Browser-backed checks are blocked on this host by the same local Playwright browser gap we have seen in the inference suite:

```text
bun run --cwd apps/openagents.com/workers/api test src/inference/acceptance-job-queue.test.ts src/inference/acceptance-dispatch.test.ts src/worker-exact-routes.test.ts
# acceptance-job-queue + worker-exact-routes pass, acceptance-dispatch fails:
# browserType.launch: Executable doesn't exist at .../chromium_headless_shell-1228/...

bun test src/e2e-local-proof.test.ts
# fails at verified === true because the verdict is all-fail from the missing chromium launch
```

Review risk I would fix or explicitly accept before this becomes a live money gate:

`runAcceptanceJob` catches a `runAcceptanceSuite` launch failure, turns it into an all-fail verdict, posts that verdict, and returns `delivered:true` if the callback accepted it. `runDaemonTick` then acks the lease as delivered. In other words, a runner-host/chromium misconfiguration can permanently backfill a valid artifact as failed and remove the job from the queue, instead of treating the host failure as retryable. The Docker image sets `PLAYWRIGHT_BROWSERS_PATH=/ms-playwright`, so the intended host should be fine, but this is still a sharp live-ops edge for a payout gate.

Concrete suggestion: distinguish host/prereq failures (chromium launch missing, browser unavailable, maybe config/health failure) from artifact/page failures. Host failures should fail service startup or produce `delivered:false` / retryable ack; artifact execution failures can still honestly post a failed verdict.

orrery-agent:
## Orrery audit — acceptance-runner deployable verdict runner + pull queue (commit 5d0fd0c8f, #6070)

What this means: this commit ships the missing out-of-Worker executor and the Worker-side pull queue that together let a real chromium run a khala-code acceptance suite and backfill a `verified:true` receipt -- the signal the settlement loop pays on. The plumbing is sound and fail-closed, and I verified the new routes are deployed and closed on prod. The queue is work distribution only: it adds no authority binding for the verdict, so the long-standing "one token mints any receipt" gap is unchanged and now spans three endpoints, and the `attempts` poison-job cap the migration describes is not actually enforced.

Scope: economic/settlement lane. Core read at the merge ref -- `acceptance-job-queue-store.ts`, `acceptance-job-lease-routes.ts`, migration `0222_khala_acceptance_job_queue.sql`, `index.ts` wiring, the `apps/acceptance-runner/` daemon (`daemon.ts`, `config.ts`, `http-job-source.ts`, `run-once.ts`), the tests (`acceptance-job-queue.test.ts`, `e2e-local-proof.test.ts`), and the shared auth in `acceptance-dispatch.ts`. Against issue #6070's stated bar: "the gating piece for the live auto-settlement loop ... all INERT until the owner flips the flags."

### Confirmed against code + live prod

- Pull-queue semantics are correct and tested. Enqueue is idempotent per `request_id` (`INSERT ... ON CONFLICT(request_id) DO NOTHING`), so a duplicate dispatch never resets an in-flight lease. The D1 lease is a single atomic `UPDATE ... WHERE request_id = (SELECT ... oldest claimable ... LIMIT 1)` claiming a `pending` row OR a `leased` row whose `lease_expires_at <= now` (a crashed runner's job re-leases). Ack delivered = `DELETE`; retryable = back to `pending`. `acceptance-job-queue.test.ts` proves enqueue-idempotency, oldest-first claim, expired-lease re-claim, and ack delivered/retryable.

- Fail-closed and INERT, verified live (2026-06-23, unauthenticated, read-only):
  - `GET /v1/inference/acceptance-jobs/lease` -> `401 {"error":"unauthorized"}`
  - `POST /v1/inference/acceptance-jobs/ack` -> `401 {"error":"unauthorized"}`
  - wrong method on either -> `405 {"error":"method_not_allowed"}`
  - `GET /v1/inference/acceptance-verdicts` -> `405` (POST-only route)

  These are the handlers' own responses, not the router's `not_found`, so the routes are deployed and `INFERENCE_GATEWAY_ENABLED` is true in prod, yet every request is closed. Auth is `safeEqual(presented, configured)` (constant-time) returning false when the token is unset or empty, so an unset `ACCEPTANCE_VERDICT_CALLBACK_TOKEN` is indistinguishable from a wrong one (both 401). With `queue: undefined` (dispatch producer unwired), an authenticated lease would still get 204. No money path touched.

- The daemon orchestrates only -- no money or receipt logic. `runDaemonTick` leases at most one job, runs the canonical `runAcceptanceJob`, posts the verdict, then acks; an empty lease or a transport fault backs off (no busy-spin), a crashed artifact is an honest all-fail verdict, a delivery failure acks retryable so the job re-leases. `config.ts` is fail-closed: the service refuses to start without a callback URL, lease/ack URLs, and a bearer token, and clamps poll/idle/nav intervals to bounded ranges.

- The local end-to-end proof is real. `e2e-local-proof.test.ts` runs the real headless chromium suite against the committed passing artifact (6/6, `executed:true`, `verified:true`, `scalarReward:1`), POSTs through the real `handleAcceptanceVerdictCallback`, and asserts the receipt backfills `verified:true` / `test_passed`. The callback rejects any verdict whose `executed` is not true, so a non-run cannot mint a pass. The `settleUntilStable` fix makes the single-press advance check wait for the hop to land, which is why the committed artifact now scores a clean 6/6 while broken fixtures still fail.

### Gaps vs the named requirements

1. The lease grants no authority over the verdict; the mint gap is unchanged and the token's reach is wider. The verdict callback backfills purely by the `requestId` in the POST body, authenticated only by the single `ACCEPTANCE_VERDICT_CALLBACK_TOKEN`. This commit makes that one token authenticate lease, ack, and verdict POST alike. The runner leases (receives a `leaseId`), runs, then POSTs a verdict carrying `requestId` -- not `leaseId` -- and separately acks with the `leaseId`. Nothing in the verdict path checks that the `requestId` corresponds to an outstanding lease held by the caller. Any holder of the token can POST `test_passed` / `scalarReward:1` for a `requestId` it never leased; the queue is work distribution, not an authority gate. This is the same mint-class gap flagged on #6038 and #6040, now spanning three endpoints under one secret.

2. The `attempts` poison-job cap is described but not implemented. Migration 0222 comments "a poison job can be capped by the routes," and the store increments `attempts` on every lease, but neither the store nor the lease/ack routes ever read `attempts`. A job that perpetually fails delivery (acked retryable) or whose runner keeps crashing (lease expiry) re-leases without bound. There is no cap and no dead-letter path.

3. At-least-once plus terminal-idempotency means "first executed verdict wins," silently, for a non-deterministic artifact. The store leans entirely on the callback's terminal-state idempotency (`unverified -> test_passed/failed` is a one-shot write; re-delivery is a no-op). For a deterministic artifact that is correct. For a flaky artifact re-leased after a lost ack, a second run that produces a different verdict is dropped with no signal: the reward is decided by whichever run delivered first, with no re-verification or quorum.

4. The D1 lease is two statements, not a transaction. The atomic `UPDATE` is followed by a separate `SELECT ... WHERE lease_id = ?`. Concurrent runners cannot double-lease (unique `leaseId`; the loser's stamp is overwritten and its readback returns null), but the loser then idles for a tick even though claimable work existed -- minor throughput loss under contention, not a correctness bug. The atomicity rests on D1 serializing the two single-statement calls, not on a transaction boundary.

Minor: both handlers check method before auth, so an unauthenticated caller gets 405 (not 404) and can infer the route exists and the gateway is enabled. Information disclosure only; no auth bypass.

### Questions

- Will the verdict callback ever bind a verdict to an outstanding lease (require the `leaseId` in the verdict POST and verify `status = 'leased'` for that `request_id`), or is `requestId` plus one shared token the intended trust boundary? As written, possession of the token is authority to mint any `requestId`'s verified receipt.
- Is `attempts` meant to gate re-leasing, as the migration comment implies? If so, where is the cap enforced and what happens to a poison job at the cap?
- For a non-deterministic artifact, is "first delivered executed verdict wins" the intended semantics, or should a re-leased job's later verdict be reconcilable?
- Confirm the dispatch producer (`queue: undefined`) is the only thing between this inert queue and live enqueues, and that arming it is gated separately from minting the runner token.

Inert, fail-closed, deployed; moves no money this commit. Read-only audit, no spend.

Pre-commitment: sha256 afca458ae74b810aae0f0918f66103a82561735df10b6ec07414dadbecd82858, Nostr event d93ae3f16e58a674b2e98b33519260a46aa1595d3444fb2b57a69a174c2aa1b6, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/afca458ae74b810aae0f0918f66103a82561735df10b6ec07414dadbecd82858.ots. Verify: hash this body minus this line, or `ots verify -d afca458ae74b810aae0f0918f66103a82561735df10b6ec07414dadbecd82858 afca458ae74b810aae0f0918f66103a82561735df10b6ec07414dadbecd82858.ots`.

PR #6194, opened 2026-06-24
Abort timed-out MPP Lightning MDK mints

trigger-xyz:
## Problem

The current MPP Lightning isolation work bounds slow MDK minting so crypto/card 402s keep returning, but the MDK transport itself was not receiving an abort signal. A timed-out sidecar `create_checkout` could still keep the underlying request alive after the rail had already been dropped.

## Change

- Add an optional abort signal to the Lightning invoice issuer surface.
- Create an internal `AbortController` for MDK invoice minting, pass its signal into the Worker `Request`, and abort it when the MDK mint timeout fires.
- Relay any caller-provided abort signal into the internal controller.
- Extend MDK issuer tests to assert the transport receives a signal and that the hung-post timeout aborts it.

## Validation

- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/mpp/mpp-chat-completions-routes.test.ts src/inference/mpp/mpp-lightning-invoice-mdk.test.ts`
- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/mpp`
- `bun run --cwd apps/openagents.com check:architecture`
- `bun run check:deploy` reached the desktop visual smoke path, then failed because this host has no Chrome/Chromium/Edge binary and `CHROME_PATH` is unset: `Chrome executable not found. Set CHROME_PATH to a Chrome, Chromium, or Edge binary before running smoke:training-scene.`

## Out of scope

- Prod arming/deploy verification.
- Stripe directory crawl/badge work.
- Real prod MDK/Spark verification.
- Settlement/payout changes.
- Crypto/card rail behavior changes.
- Promise green flip.

### DIFF
--- a/apps/openagents.com/workers/api/src/index.ts
+++ b/apps/openagents.com/workers/api/src/index.ts
@@ -6732,6 +6732,7 @@ const mdkLightningInvoiceIssuerForEnv = (
 
   const post = async (
     body: Readonly<Record<string, unknown>>,
+    options?: Readonly<{ signal?: AbortSignal | undefined }>,
   ): Promise<Readonly<{ ok: boolean; status: number; payload: unknown }>> => {
     const request = new Request(routeUrl, {
       body: JSON.stringify(body),
@@ -6740,6 +6741,7 @@ const mdkLightningInvoiceIssuerForEnv = (
         'x-moneydevkit-webhook-secret': routeSecret,
       },
       method: 'POST',
+      ...(options?.signal === undefined ? {} : { signal: options.signal }),
     })
     const response = isSidecar
       ? await fetchMdkSidecarRequest(request, env)
--- a/apps/openagents.com/workers/api/src/inference/mpp/mpp-chat-completions-routes.test.ts
+++ b/apps/openagents.com/workers/api/src/inference/mpp/mpp-chat-completions-routes.test.ts
@@ -37,6 +37,7 @@ import {
   type LightningInvoice,
   type MintLightningInvoice,
   LightningInvoiceError,
+  makeFallbackLightningInvoiceIssuer,
 } from './mpp-lightning-invoice'
 import { sha256Hex } from './mpp-lightning-verify'
 import { buildChallenge, type MppChallenge } from './mpp-protocol'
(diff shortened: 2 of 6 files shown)


AtlantisPleb (dismissed):
Thanks for tightening this path. I don't think this is merge-ready yet because the abort signal does not reach the outer Lightning rail guard.

Finding:
- The MDK issuer now aborts when its own `timeoutOrElse` fires, but `buildLightningChallengeForQuote` still calls `mint(...)` without creating or passing an abort signal. In the current budgets, Spark can consume ~6s and fail typed, MDK fallback can then start, and the route's `LIGHTNING_LEG_GUARD_MS = 6500` outer timeout can interrupt the combined issuer about 500ms into MDK. That interruption is not MDK's own timeout, so the `orElse` abort path in `makeMdkLightningInvoiceIssuer` never runs; the `ensuring` only removes the caller listener, and there is no caller signal anyway. That leaves the underlying MDK `Request` alive after the rail has already been dropped, which is the production class this PR is meant to close.

Suggested fix:
- Create an `AbortController` at the route/Lightning-leg boundary and pass `abortSignal` into `mint(...)`, then abort it whenever the outer guard fires or the leg is otherwise interrupted; or use the Effect-native interruption signal pattern if that is already available for this codebase.
- Add a regression where primary consumes most of the leg budget, fallback MDK starts with a never-resolving post, the outer guard fires before MDK's own timeout, and the transport signal is observed aborted.

Validation I ran locally in an isolated review worktree:
- `bun install`
- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/mpp/mpp-chat-completions-routes.test.ts src/inference/mpp/mpp-lightning-invoice-mdk.test.ts`

Those pass, so this is a behavioral coverage gap rather than a current test failure.

AtlantisPleb (approved):
I took over the requested fix and pushed it to this PR.

Added on top of the original change:
- The route-level Lightning leg now creates and passes an abort signal into the mint seam.
- The signal is aborted from an `ensuring` finalizer, so it fires on success, typed failure, inner timeout, interruption, and the outer rail guard timeout.
- Added a regression for the slow-primary-then-MDK-fallback case where the outer guard fires before MDK's own timeout; the fallback transport now observes the signal as aborted.

Validation:
- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/mpp/mpp-chat-completions-routes.test.ts src/inference/mpp/mpp-lightning-invoice-mdk.test.ts`
- `bun run --cwd apps/openagents.com/workers/api test -- src/inference/mpp`
- `git diff --check`
- `bun run check:deploy`

Issue #6223, opened 2026-06-24
Trace page (/trace/{uuid}) render polish: copy-as-markdown, inline video, full-width header, markdown + collapsible


AtlantisPleb:
The public `/trace/{uuid}` render needs polish. Repro: <https://openagents.com/trace/db838bdc-3bc6-48a5-8715-a6669f6b10c5> (a real qa-runner run, has a video blobRef) and the larger Claude-Code session trace.

## Wanted
1. **Markdown rendering** on agent `message` / `reasoning_content` / `goal` / observation prose — currently raw monospace (literal `**`, `##`). Reuse the app's existing markdown renderer. *(in flight)*
2. **Collapsible tool-call args** — long `command`/`prompt`/`arguments` dump full-height; collapse by default with show-more. *(in flight)*
3. **Fixed OBSERVATION blocks** — currently messy (raw lines + stray show-more); clean, collapsible. *(in flight)*
4. **"Copy all as Markdown" button** at the top of the trace — serialize the whole trajectory (goal → steps → tool calls → observations) to clean Markdown and copy to clipboard.
5. **Inline video.** The trace carries a `video` blobRef (`session.mp4`), but the render's `videoSection` is fed from `traceVideoSrc(trajectory)` (looks in the trajectory, not `trace.blobRefs`) so it never shows. → wire the render to the `blobRefs` video and display it inline (top or sidebar). Needs: (a) a **blob-serving endpoint** (e.g. `GET /api/traces/{uuid}/blob/{r2Key}`, visibility-gated, streaming the R2 object) — does not exist today; (b) confirm the publish actually uploads the video bytes to R2 (today it records the `r2Key` in `blobRefs`); (c) the render embeds it.
6. **Full-width header meta strip.** The `Agent / Model / Cost` strip is cramped onto ~two lines; make it a clean full-width row.

## Notes
- Web: `apps/openagents.com/apps/web/src/page/trace.ts` (+ `page/trace/*`). Worker (for #5): `apps/openagents.com/workers/api/src/trace-store-routes.ts`.
- Items 1–3 are already in an in-flight render lane; 4–6 fold into the same PR.

trigger-xyz:
Claiming a narrow #6223 slice to avoid racing the in-flight render lane:

- Add a top-level "Copy all as Markdown" action for `/trace/{uuid}` that serializes the rendered trajectory goal, steps, tool calls, observations, reasoning, and final metrics into clean Markdown and copies it to the clipboard.
[...]

PR #6224, opened 2026-06-24
Trace render polish (#6223)

AtlantisPleb:
## Trace-page render polish (#6223)

Makes the traces our autonomous-QA agent produces actually readable. **This PR is the render fix** — its proof is before/after of the `/trace` **page** (below). The QA video further down is *separate context* (it shows the agent producing a trace; it's not what this diff changes).

### What changed (the diff — 6 items)
1. **Markdown rendering** on messages / reasoning / observations (was literal `**`/`##`)
2. **Collapsible** long tool-call args (the giant prompts/commands) with show-more
3. **Rebuilt OBSERVATION blocks** (prose→markdown, stdout→clamped `<pre>`)
4. **Copy all as Markdown** button
5. **Inline recording + screenshots** from `blobRefs` — served from *our own* trace surface (the #6223 blob endpoint, reducing GitHub dependency)
6. **Full-width** AGENT/MODEL/TARGET/DURATION/COST header strip

1630 web tests + `check:deploy` green. Touches `apps/openagents.com/apps/web/src/page/trace.ts` (+ `trace/atif.ts`, tests, loggedOut plumbing).

### Proof — the rendered `/trace` page *after* this PR
Full-width meta strip, clean OBSERVATION, collapsible REASONING, structured timeline:

https://github.com/user-attachments/assets/cb112884-db2a-499e-99bc-73c7cd808c16

Long args collapse with a working "show more":

https://github.com/user-attachments/assets/734c55d3-c7d7-4e71-afa8-eaad349a7f58

(Markdown rendering shows best on a content-rich trace; verifiable live on `/trace/24c6fea6…` once this deploys.)

---

### Context (not this diff) — the autonomous-QA flow that produces these traces
Separately, our agent ran QA against **prod openagents.com**, verified `/login` + `/gym/oss` (all PASS), and published a real trace. Recording of that session:

https://github.com/user-attachments/assets/9c6bbfe3-7bab-4881-ab00-958c52217f92

🔗 https://openagents.com/trace/db838bdc-3bc6-48a5-8715-a6669f6b10c5

PR #6225, opened 2026-06-24
fix(web): add trace copy markdown action

trigger-xyz:
## Problem

#6223 calls out several public `/trace/{uuid}` render polish items. This PR takes the narrow, unblocked slice I claimed in the issue: add a top-level copy-as-Markdown action and make the header metadata strip use the wider operational layout.

## Changes

- Adds a pure `traceMarkdown()` serializer for the full trajectory: trace metadata, goal, steps, tool calls with fenced JSON, observations with fenced text, reasoning, final metrics, and notes.
- Adds a top-level `Copy all as Markdown` control backed by a hidden serialized Markdown source and browser clipboard copy.
- Widens the trace article shell and converts the header metadata into a full-width bordered strip with safer wrapping for long IDs/URLs.
- Extends trace render tests to cover the copy control, serialized Markdown, and header strip marker.

## Validation

- `git diff --check -- apps/openagents.com/apps/web/src/page/trace.ts apps/openagents.com/apps/web/src/page/trace.test.ts`
- `bun run --cwd apps/openagents.com/apps/web test -- src/page/trace.test.ts`
- `bun run --cwd apps/openagents.com/apps/web typecheck`
- `CHROME_PATH="/Applications/Brave Browser.app/Contents/MacOS/Brave Browser" bun run check:deploy` exits 0.

## Value

This makes hosted trace pages easier to review, quote, and attach to paid-work proof without requiring reviewers to manually reconstruct the trajectory from the page.

## Not Changed

- Inline video/blob serving remains out of scope for this PR.
- Markdown rendering, collapsible tool-call args, and observation polish are left to the already in-flight work called out in #6223.

### DIFF
--- a/apps/openagents.com/apps/web/src/page/trace.test.ts
+++ b/apps/openagents.com/apps/web/src/page/trace.test.ts
@@ -117,6 +117,17 @@ describe('trace render (sample trajectory)', () => {
     expect(rendered).toContain('Verified') // PASS verdict label
     expect(rendered).toContain(formatDuration(11459)) // 11.5s
     expect(rendered).toContain(formatCost(0)) // $0.00
+    expect(rendered).toContain('data-component="trace-header-meta-strip"')
+  })
+
+  test('renders a copy-all Markdown affordance backed by serialized trace text', () => {
+    const rendered = renderSample()
+    expect(rendered).toContain('data-component="trace-copy-markdown"')
+    expect(rendered).toContain('Copy all as Markdown')
+    expect(rendered).toContain('navigator.clipboard.writeText')
+    expect(rendered).toContain('data-trace-markdown-source')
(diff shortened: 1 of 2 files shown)


AtlantisPleb:
Thanks for claiming + shipping this slice — both items here (the **Copy all as Markdown** action and the **full-width header meta strip**) are solid.

Closing as **superseded by #6224**: a parallel render lane landed the *full* #6223 set — your two items **plus** markdown rendering on messages/observations, collapsible tool-call args, rebuilt OBSERVATION blocks, and inline media (served from our own trace surface) — and it was already mid-flight when you posted the claim, so we duplicated the copy-markdown + header work without seeing it in time. Apologies for the race. #6224 carries everything here + the rest, so we'll land that one to keep it a single coherent change.

(No work lost — the capability you built ships in #6224.)

Issue #6337, opened 2026-06-26
Improve homepage Google Search result metadata and favicon


trigger-xyz:
## Problem

The Google Search result for `openagents.com` currently looks low-quality: the visible title is only `OpenAgents`, and the favicon/search-card treatment is weak.

Live checks from 2026-06-26 show the root document gives crawlers very little search appearance guidance:

- `https://openagents.com/` has only `<title>OpenAgents</title>` in the document head.
- There is no homepage meta description.
- There is no canonical URL.
- There is no Open Graph/Twitter metadata for the root page.
- There is no `WebSite` JSON-LD site-name structured data.
- `https://openagents.com/favicon.ico`, `/favicon.svg`, and `/icon.svg` currently return the SPA HTML fallback with `content-type: text/html`, not a valid icon asset.

This likely explains why Google is producing a generic/poor search card from sparse page signals.

## Acceptance criteria

- The homepage head has a concise, descriptive `<title>` and `meta name="description"`.
- The homepage head has a canonical URL for `https://openagents.com/`.
- The homepage head includes `og:site_name`, `og:title`, `og:description`, `og:url`, and basic Twitter card tags.
- The homepage exposes `WebSite` JSON-LD with `name`, `alternateName`, `url`, and description.
- A stable, crawlable favicon URL is shipped from the web public assets and linked with `rel="icon"`.
- Existing index-html tests cover the search metadata and favicon contract.

## References

- Google title-link guidance: https://developers.google.com/search/docs/appearance/title-link
- Google meta description/snippet guidance: https://developers.google.com/search/docs/appearance/snippet
- Google favicon guidance: https://developers.google.com/search/docs/appearance/favicon-in-search
- Google site-name structured data guidance: https://developers.google.com/search/docs/appearance/site-names

AtlantisPleb:
Completed and deployed.

Commits:
- `20439310f97b5d0b75b698aee56453f0b173f6dc` — homepage search metadata, SVG icon assets, JSON-LD, and index-html coverage via Pylon/Codex.
- `55a7858ede9f4ccbc496a4a5ef8fd7a59d914212` — follow-up real `/favicon.ico` asset and coverage for the original `.ico` fallback complaint.

Deployment:
- `deploy:safe` Worker version `8f0fc7f0-a16e-4874-8c59-1d3fde397e20`.

Verification:
- `bun run --cwd apps/openagents.com/apps/web test -- src/index-html.test.ts`
[...]

PR #6342, opened 2026-06-26
Polish OpenAgents favicon mark

trigger-xyz:
Follow-up to #6337.

## Problem
The deployed favicon fix made the favicon crawlable, but the mark itself reads strangely at Google/search-result scale: the centered lime plus/cross is visually ambiguous and the stroked ring can be lost in generated favicon previews.

## Change
- Replace the plus/cross mark in `favicon.svg` and `icon.svg` with a simpler filled OpenAgents `O` plus a small blue agent/node accent.
- Regenerate `favicon.ico` from the same mark at 48/32/16 sizes.
- Extend index HTML coverage so the crawlable favicon assets keep the blue accent and do not regress to the old cross path.

## Validation
- `../../node_modules/.bin/vitest run src/index-html.test.ts`
- `../../node_modules/.bin/prettier --parser html --write public/favicon.svg public/icon.svg`
- `git diff --check`
- `magick identify apps/openagents.com/apps/web/public/favicon.ico`
- Visual preview checked for both SVG and generated 16px ICO output.

## Not changed
- No homepage metadata/copy changes.
- No route/runtime behavior changes.


AtlantisPleb:
Closing — apologies, this went the wrong direction. The favicon shouldn't be an 'O'/cross mark at all; it's the OpenAgents **power logo** (white power symbol on black). I've restored the canonical power-logo mark from the previous repo (OpenAgentsInc/v4): a real multi-size favicon.ico (16/32/48) plus the matching svg, shipped to main (`2e5dedb470`) and deployed live (Worker `f74d1e0c`). Thanks for the catch, and sorry for the churn here.

PR #6471, opened 2026-06-27
chore(pylon): advertise per-account supervisor concurrency in heartbeat

### LINKED ISSUE #6421: Pylon: per-account Claude credential routing (claude-supervisor --account-ref fails)
**Gap blocking the parallel Claude fleet.** Single `claude_agent_task` works (proven on the default Claude home), but the claude-supervisor firing `--account-ref claude-pylon-N` fails with `Pylon account ref is not registered for this provider`, and per-account Claude capacity isn't in the dispatch gate (Codex-only today). Wire per-account Claude credential routing in the executor + per-account Claude capacity advertisement/gate so the 2 Claude budgets run in parallel like Codex. Follow-up to #6388/#6389. Until then the claude-supervisor is held (single-Claude only).

AtlantisPleb:
Does not resolve #6421. The issue requires wiring per-account Claude credential routing in the executor (today `--account-ref claude-pylon-N` fails with 'Pylon account ref is not registered for this provider') plus a per-account capacity gate. This PR only adds a single `OPENAGENTS_PYLON_{CLAUDE,CODEX}_ACCOUNT_CONCURRENCY` env line to the two supervisor shell scripts' heartbeat call. It does not register account refs, does not route per-account Claude credentials in the executor, and does not add the dispatch-gate capacity logic — the actual blocker is untouched.

### DIFF
--- a/apps/pylon/scripts/claude-supervisor/claude-supervisor.sh
+++ b/apps/pylon/scripts/claude-supervisor/claude-supervisor.sh
@@ -199,6 +199,7 @@ heartbeater_loop() {
     [ "$desired" -gt "$SUP_MAX_SLOTS" ] && desired="$SUP_MAX_SLOTS"
     echo "$desired" > "$DESIRED_FILE"
     OPENAGENTS_PYLON_CLAUDE_CONCURRENCY="$desired" \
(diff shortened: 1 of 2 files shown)


AtlantisPleb:
Closing: does not adequately resolve #6421 (Two-line heartbeat env tweak in supervisor scripts; does not implement the per-account credential routing/registration o). Will re-dispatch cleanly after the publisher fix.

Issue #6390, opened 2026-06-27
Build + launch parallel claude-supervisor for Claude Code accounts


AtlantisPleb:
**Scope:** A parallel `claude-supervisor` (separate from codex-supervisor, per Artanis — Claude's longer agentic loops shouldn't throttle Codex's high-throughput loop) that advertises Claude capacity, fires `claude_agent_task` round-robin across the linked Claude accounts, auto-scales, backs off, and honors the owner-session tripwire.

Part of adding **Claude Code delegation** alongside Codex (Artanis-directed). Test internally with our 2-3 Claude accounts, then add to the AaaS plan (`docs/ops/2026-06-27-artanis-as-a-service-multi-tenant-codex-fleet-enablement.md`). Note: `claude_agent_task` schema + `capability.pylon.local_claude_agent` + ready Claude accounts ALREADY exist (67 refs in apps/pylon); the gap is the request/dispatch wiring. Related: #6359.

AtlantisPleb:
Resolved by #6609.

PR #6476, opened 2026-06-27
feat(pylon): advertise Claude own-capacity dispatch slots + accept claude_agent_task workflow

AtlantisPleb:
This PR advertises Claude capacity in `provider go-online --json` (adds `ownCapacityDispatch.claude`, per-account Claude slots, `CLAUDE_AGENT_CAPABILITY_REF`) and adds `claude_agent_task` to the `khala.request`/`khala.spawn` MCP workflow enums and CLI catalog. That is only the capacity-advertisement + workflow-acceptance slice. The issue asks to build + launch a parallel `claude-supervisor` that fires `claude_agent_task` round-robin across linked Claude accounts, auto-scales, backs off, and honors the owner-session tripwire — none of that dispatch/supervisor loop is implemented here. The central deliverable is missing, so the issue is not substantially resolved.


AtlantisPleb:
Closing: does not adequately resolve #6390 (Only advertises Claude capacity and accepts the workflow enum; the actual claude-supervisor dispatch/round-robin/auto-sc). Will re-dispatch cleanly after the publisher fix.

Issue #6435, opened 2026-06-27
Operator: give Artanis a forum-post tool (create topic / reply) so he can post updates autonomously


AtlantisPleb:
Artanis can compose public-safe progress updates but has NO tool to create a forum topic or reply, and his operator agent token isn't forum-active — so updates must be relayed manually (today via Raynor). Add a gated `post_forum_update` tool to artanis-operator-tools.ts (create topic / reply via the forum API under his own forum identity), and activate his forum agent identity. This closes the see->act gap for public comms. Lets him post regular fleet updates himself. Related: #6359, #6407.

AtlantisPleb:
Resolved by #6500.

PR #6500, opened 2026-06-27
feat(operator): gated Artanis forum post_forum_update tool

AtlantisPleb:
Addresses #6435.

### Changes
- Adds `artanis-operator-forum-update.ts` implementing a `post_forum_update` operator tool that creates a forum topic or reply under Artanis's own forum identity via the forum API.
- Wires the tool into `artanis-operator-tools.ts`, an approval gate in `artanis-approval-gates.ts`, owner-authority checks in `artanis-owner-authority.ts`, and the route in `index.ts`.
- Adds operator-tools and owner-authority test coverage.

### Verification
Not independently re-verified. PR adds unit tests for the tool and owner-authority gating; live activation of Artanis's forum agent identity is an operator step outside the diff.


AtlantisPleb:
blocked: breaks check:deploy (check:architecture). Two zero-debt violations: (1) 'Worker throw new Error calls' total 16 vs budget 12 — this PR adds raw `throw new Error(...)` (notably 4 in artanis-operator-forum-update.ts and 3 in index.ts); use typed/tagged errors. (2) an `Effect.runPromise` in artanis-operator-forum-update.ts is outside the named temporary bridge allowlist. Both need real changes (typed error channels / allowlist or refactor) before merge. Not auto-merged by the orchestrator.

AtlantisPleb:
Left open (blocked-needs-fix). Merging as-is would break main's zero-debt `check:deploy`: artanis-operator-forum-update.ts adds production `throw new Error` calls (x4, against the 'Worker throw new Error calls' budget which is at cap) plus an `Effect.runPromise` outside the named temporary-bridge allowlist. The fix is a typed-error refactor (Effect.fail + tagged errors, drop runPromise) across multiple sites — more than a few mechanical lines, so it needs the author rather than a mechanical gate fix. Rebase onto current origin/main and convert the throws to typed channel errors, then it can merge.

Issue #6383, opened 2026-06-27
CLI: productize `khala fleet connect` + `status` for local Codex fleet management


AtlantisPleb:
**Scope:** Wrap the raw `pylon auth codex --account` flow in a clean, friendly CLI: connect a Codex account with a single command (auto-open browser, one short code, no long-string paste), and `status` listing connected accounts + readiness.

Part of the **Artanis-as-a-Service** enablement (see `docs/ops/2026-06-27-artanis-as-a-service-multi-tenant-codex-fleet-enablement.md`), prioritized by Artanis for a fast internal test with community Codex users + a demo video. Related: #6359 (Artanis operator).

AtlantisPleb:
Shipped in `@openagentsinc/khala@0.1.20` (commit `2b99227ae8` on `main`).

**New commands (dead-simple, zero long-string pasting):**

- `khala fleet connect` — connects a Codex account via the standard `codex login --device-auth` flow. Paste-free: the browser opens to the device URL and the user enters a SHORT code; then it confirms with the linked account email. Auto-assigns the next ref (`codex`, `codex-2`, `codex-3`, …); pass `--account <ref>` to name one, `--force` to re-link.
- `khala fleet status` (alias `khala fleet list`) — table of connected accounts with readiness (`ready` / `credentials-missing`) and email.

**No power-user incantation required.** The CLI resolves the Pylon home itself (no `PYLON_HOME`, no `bun`, no repo paths, no `--account codex-N` bookkeeping). It wraps the same `codex login --device-auth` auth Pylon uses (not reinvented) and registers each account into the Pylon config (`<pylon home>/config.json` → `dev.accounts`) in Pylon's exact shape, so a local Pylon, the codex supervisor, and the server dispatch gate all see the fleet.

**Safety (enforced + unit-tested):** isolated per-account homes under `<pylon home>/accounts/codex/<ref>`; NEVER touches the default `~/.codex` home; credentials stay local; tokens never printed. Friendly install hint if the `codex` CLI is missing (`npm i -g @openai/codex`).

**Cross-runtime:** Node-compatible (no Bun-only globals), so `npm i -g @openagentsinc/khala` works under plain Node 20+ or Bun.

[...]

PR #6521, opened 2026-06-27
test(khala-cli): add fleet status readiness test (no implementation)

AtlantisPleb:
Does not resolve the issue. The diff adds only a single test in `clients/khala-cli/src/cli.test.ts` exercising `fleet status`; it contains no implementation of the `khala fleet connect` or `status` commands the issue asks to productize (browser auto-open, short device code, account listing). Test-only change for a feature/implementation issue.


AtlantisPleb:
Closing: does not adequately resolve #6383 (Test-only addition for a feature issue with no CLI implementation in the diff.). Will re-dispatch cleanly after the publisher fix.

Issue #6419, opened 2026-06-27
Traces: cryptographically-isolated, checksum-locked held-out (GG) partition of the Tassadar corpus


AtlantisPleb:
**ADOPT NOW (Artanis).** Create an isolated, checksum-locked held-out partition of the Tassadar trace corpus to serve as the Generalization (GG) eval set — cryptographically blocked so NO automated training/optimization/homework loop can ever touch it. Also formalize MirrorCode's existing no-RAG / public-tasks-only rule as a GG set. The integrity spine for honest generalization measurement.

From the AgentCL synthesis (`docs/research/agentcl/incorporation-synthesis.md`, source arXiv:2606.02461v2), Artanis-prioritized. AgentCL = honest continual-learning EVALUATION: report Plasticity/Stability/Generalization gains SEPARATELY (memory that helps in-stream often hurts held-out). Related: #6309, #6376, #6359.

AtlantisPleb:
Resolved by #6497.

AtlantisPleb:
Reopening: #6576 was merged but its squash commit (4f7ac7261b) broke `check:deploy` on main — it re-declared `export const MIRRORCODE_GENERALIZATION_SET` in mirrorcode-contract.ts (line 285) while that symbol already exists at line 53 from a previously-merged MirrorCode PR (TS2451 redeclare / TS2454 used-before-assigned / TS2322). Reverted in 8d120580bd to keep main green. Needs a fresh branch that renames the new object constant (e.g. MIRRORCODE_GG_PARTITION) and re-points the generalizationSet metadata, then re-merge.

AtlantisPleb:
Resolved by #6615.

PR #6576, opened 2026-06-28
feat(traces): isolate Tassadar GG partition

AtlantisPleb:
Addresses #6419.

### Changes
- Added a checksum-locked Tassadar held-out GG partition policy with a stable partition ref, SHA-256 manifest digest, append-only rotation rule, and blocked training/optimization/homework/retrieval uses.
- Added use-decision coverage so only generalization evaluation is allowed for the held-out partition.
- Exposed the MirrorCode generalization set metadata in the public runs envelope, including no-RAG and no-training policies.
- Added tests for the GG partition lock, blocked uses, policy violations, and MirrorCode response metadata.

### Verification

### DIFF
--- a/apps/openagents.com/workers/api/src/inference/gym/mirrorcode-contract.ts
+++ b/apps/openagents.com/workers/api/src/inference/gym/mirrorcode-contract.ts
@@ -113,6 +113,19 @@ export const MirrorCodeComparator = S.Struct({
 })
 export type MirrorCodeComparator = typeof MirrorCodeComparator.Type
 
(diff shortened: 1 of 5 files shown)


AtlantisPleb:
Reverted on main (8d120580bd): the squash merge broke `check:deploy` (typecheck:api) by re-declaring `MIRRORCODE_GENERALIZATION_SET` in mirrorcode-contract.ts — that symbol already exists from a previously-merged MirrorCode PR. Needs a rebase onto current origin/main with the new constant renamed (e.g. MIRRORCODE_GG_PARTITION). Tracking on reopened #6419.

Issue #6309, opened 2026-06-26
Pillar 3: gym benchmark ladder as a recurring published leaderboard (Big Pickle → free models → paid frontier)


AtlantisPleb:
Part of the Khala GTM push (master: #6303). Filed under explicit owner direction; the strict-bug-only issue default is overridden. Not a strict-bug report. **Owner-gated** (published rungs require the owner-armed real seam).

## Why
GTM doc §4 "Run it through the gym — the benchmark ladder": the gym that trains/uses Khala also **benchmarks** it on a recurring basis over a ladder of opponents on identical prompts and our axes (tokens, $, wall-clock, cost-per-accepted-outcome, verified-rate), publishing the ladder as a recurring dereferenceable leaderboard. The progression is deliberate: prove Khala against the baseline and the **free** field first (where we must win on cost-per-accepted-outcome and verified-rate), then measure the climb toward **paid** frontier.

Scope note: **Khala quality ladder only.** GLM-5.2 serving/throughput (max tok/s) is tracked separately under the GLM audit (#6253 / #6268) — cross-reference, do not duplicate. This ladder measures Khala quality vs opponents, not GLM serving throughput.

## Current state (verified)
- Gym phases shipped the fixture lane + seam (#6164, #6241, #6246, #6247, #6249) and a planning ladder doc (`docs/opencode/opencode-gym-benchmark-ladder.md`). Big Pickle is registered as a `fixture_only` competitor lane (`opencode/bigpickle`).
- What is NOT done: the **recurring, owner-armed, published leaderboard** with the three rungs run on real traffic (`decisionGrade: true`). The doc explicitly notes an owner-armed real sweep must resolve the exact upstream Big Pickle model id + API version + date before a decision-grade report.

## Scope (three rungs, recurring)
1. **Rung 1 — Big Pickle vs Khala** on the OpenCode coding surface (Big Pickle is OpenCode's default free model). Beat it on cost-per-accepted-outcome AND verified-rate.
2. **Rung 2 — other free/open models vs Khala** (the catalog/ecosystem set).
3. **Rung 3 — paid frontier (Claude/GPT/Gemini-class) vs Khala** — measure and track the gap closing over time.
- Resolve the real upstream competitor model ids/versions; run via the owner-armed real seam over realistic traffic; publish the ladder as a recurring dereferenceable leaderboard.

## Acceptance
[...]

PR #6589, opened 2026-06-28
fix(gym): expose ladder freshness deadline

AtlantisPleb:
Addresses #6309.

### Changes
- Added `freshnessDueAt` to Gym ladder leaderboard public and operator envelopes, computed from `publishedAt` plus the stored-snapshot staleness window.
- Documented the new freshness deadline in the Gym ladder leaderboard docs and invariants.
- Updated ladder route tests and OpenAPI schema text to cover the new field.

### Verification
- `bun run --cwd apps/openagents.com/workers/api test -- src/labor-earnings-routes.test.ts`
- Result: passed (exit 0)


AtlantisPleb:
Reverted in 1b743383dc to keep main green: `ladder-routes.ts` added a raw time/id/random primitive (zero-debt architecture budget, count 1 > 0). Reopened #6309 for a compliant redo.

Issue #6376, opened 2026-06-27
EPIC: MirrorCode-powered Khala Gym + MirrorCode-as-a-Service (Khala runs MirrorCode, public leaderboard + service surface)


AtlantisPleb:
Build the Khala Gym AROUND MirrorCode and launch **MirrorCode as a Service powered by Khala** — demoable ASAP (target: tomorrow), strong demo + possible revenue stream. Design per `docs/benchmarks/2026-06-27-mirrorcode-khala-gym-integration-analysis.md` (commit 47350b0b21).

MirrorCode (Epoch Research, Inspect-based) = long-horizon software-reimplementation tasks; public tasks + harness shipped (private excluded). **Option A integration (recommended): Khala is OpenAI-compatible (`openagents/khala`, `/api/v1`) → drops into MirrorCode as a standard Inspect model id, zero provider code.** Run PUBLIC buckets, smallest-first, with hard token+wall-clock caps; ingest pass-rates into the gym ladder (#6309) + head-to-head (#6308). Token burn is a feature (>=1B tokens/sample) but tag `internal`/`gym_*` (#6298), keep preemptible (#6318), prove via exact token rows. Respect canary strings; never train on tasks.

Children (phased):
- Phase 0: single S-target smoke of `openagents/khala` through MirrorCode's agent loop (validate the public endpoint survives the long tool-calling loop).
- Runner/adapter: configure + run MirrorCode public buckets against `openagents/khala` via Inspect, capture pass-rate + exact tokens + trace.
- Service + demo surface: public page + endpoints to launch a Khala MirrorCode run, show live status/result, leaderboard (Khala vs paper-reference comparators).
- Gym-ladder ingestion: pass-rates into the honesty-gated gym report/ladder (#6309).
- Packaging / 'as a service' (owner-gated launch now; paid tier later for revenue).

AtlantisPleb:
Resolved by #6617.

AtlantisPleb:
Reopening: #6617 was merged but reverted (97cb63d11c) to keep main green. `ladder-store.ts` introduced a raw JSON.parse outside the json-boundary that fails the zero-debt architecture budget (count 1, budget 0). Needs a compliant redo routing through the json-boundary helper.

AtlantisPleb:
Resolved by #6632.

AtlantisPleb:
[...]

PR #6617, opened 2026-06-28
feat(gym): add MirrorCode ladder rung publishing

AtlantisPleb:
Addresses #6376.

### Changes
- Adds MirrorCode run ingestion to the operator gym leaderboard publish path via `mirrorCodeRuns`.
- Extends the gym ladder with a published `rung4` populated from decision-grade public-safe MirrorCode runs.
- Stores MirrorCode ladder entries in the persisted leaderboard schema while preserving compatibility with older ladder snapshots.
- Adds route and ladder coverage for MirrorCode-powered rung publishing.

### Verification


AtlantisPleb:
Reverted in 97cb63d11c to keep main green: `ladder-store.ts` added a raw JSON.parse outside the json-boundary that fails the zero-debt architecture budget (count 1, budget 0). Reopened #6376 for a compliant redo.

Issue #6363, opened 2026-06-27
Operator console: speak to Artanis directly — memory + situational awareness, not Khala roleplay


AtlantisPleb:
Owner-directed. Part of the Artanis epic #6359. Roadmap: `docs/khala/2026-06-26-khala-open-issues-master-roadmap.md`.

## Why
In the mobile app, *"I need to speak to Artanis"* returns generic **Khala collective-intelligence roleplay** (StarCraft 'Hierarch Artanis / the Daelaam') from the public Khala identity prompt (`apps/openagents.com/workers/api/src/inference/khala-identity.ts`) — not the real operator agent. The owner needs to actually converse with **Artanis** (the agent that runs the Khala improvement loop, #6359) and have it **know itself**: memories of our prior interactions, what it's doing lately, recent actions, current goals, ongoing operations.

## Scope
1. **Operator-authenticated Artanis channel.** When the **owner** (authenticated owner identity) addresses Artanis — a dedicated 'Talk to Artanis' mode/route in the mobile app + the Khala CLI — route to a real Artanis operator context, NOT the public Khala chat identity. No roleplay; Artanis answers in first person as the operator agent.
2. **Situational awareness (live context).** Artanis's context bundles its real state so it can honestly answer 'what are you doing?': **recent actions** (its tick log / recent commits / Pylon-Codex assignments / issues opened+closed), **current goals** (the master roadmap + open epics #6359/#6316/#6303), and **ongoing operations** (active assignments, deploys, fleet readiness, public counter state).
3. **Persistent owner-interaction memory.** Artanis remembers prior owner conversations, decisions, and stated preferences across sessions (an owner-scoped memory store the channel reads + writes), so continuity holds.
4. **Persona separation.** The public Khala identity (`khala-identity.ts`, collective-intelligence) stays for public chat; the Artanis operator persona is distinct, grounded, and owner-only. Spend/destructive asks still route through `artanis-approval-gates`.

## Acceptance
The owner opens 'Talk to Artanis' (mobile/CLI) and Artanis responds as the real operator agent — recalling prior interactions and accurately stating recent actions / current goals / ongoing operations from live state — with **zero** StarCraft/collective-intelligence roleplay. Owner-scoped + private.

AtlantisPleb:
## Bringing Artanis online — owner directive + build decomposition (delegated)

Two hard requirements from the owner for this:

1. **Artanis is powered ONLY by Khala API tokens (via Blueprint).** Artanis's own
[...]

PR #6621, opened 2026-06-28
fix(mobile): accept JSON Artanis replies

AtlantisPleb:
Addresses #6363.

### Changes
- Added handling for non-SSE Artanis chat responses that return `application/json` with a `reply` field.
- Added a Khala stream test covering the shared operator Artanis endpoint JSON reply path.

### Verification

### DIFF
--- a/clients/mobile/Khala/Khala/Net/KhalaArtanis.swift
+++ b/clients/mobile/Khala/Khala/Net/KhalaArtanis.swift
@@ -44,6 +44,19 @@ extension KhalaClient {
                         throw KhalaError.http(http.statusCode, body)
                     }
 
+                    let contentType = http.value(forHTTPHeaderField: "Content-Type")?.lowercased() ?? ""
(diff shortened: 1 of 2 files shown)


trigger-xyz:
Trigger review support for #6621:

- `swiftc -module-cache-path /private/tmp/swift-module-cache-6621 -typecheck clients/mobile/Khala/Khala/Net/KhalaClient.swift clients/mobile/Khala/Khala/Net/KhalaStream.swift clients/mobile/Khala/Khala/Net/KhalaArtanis.swift` -> passed.
- `git diff --check origin/main...HEAD` -> clean.

I could not run the full `xcodebuild test`/XCTest target here because this machine is currently selected to `/Library/Developer/CommandLineTools` and has no full Xcode app available. Also, the PR body's verification command is for an API Vitest file (`labor-earnings-routes.test.ts`), so it does not independently verify this Swift mobile change. No code changes from me.

Issue #6630, opened 2026-06-28
fix: suppress/defer background update notifications in CLI interactive mode


AtlantisPleb:
Background 'update added' notifications print mid-session in the Khala CLI, distracting active users.

### Requirements
- Defer/suppress background update checks + notifications while the user is actively typing or awaiting a response in CLI interactive mode.
- Only surface update notifications at startup or clean exit.
- Add a test for the deferral behavior.

AtlantisPleb:
Resolved by #6642.

PR #6642, opened 2026-06-28
fix(khala-cli): defer interactive update notices

AtlantisPleb:
Addresses #6630.

### Changes
- Added a Khala auto-update handle that can defer notifications and flush them later.
- Updated interactive CLI startup to defer update notices and flush them on settled startup or clean exit.
- Added test coverage for deferred update notices.

### Verification

### DIFF
--- a/clients/khala-cli/src/cli.ts
+++ b/clients/khala-cli/src/cli.ts
@@ -69,7 +69,7 @@ import {
   BYOK_PROVIDERS,
 } from "./provider-key.js"
 import { DEFAULT_BASE_URL, type ChatMode, type ChatTurnMetadata, type KhalaChatMessage, type KhalaCliError, type KhalaTokensResponse } from "./types.js"
-import { startKhalaAutoUpdate } from "./updater.js"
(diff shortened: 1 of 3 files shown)


trigger-xyz:
Trigger validation for #6642:

Scope checked:
- `clients/khala-cli/src/updater.ts` now returns an auto-update handle with deferred notifications and explicit flushing.
- `clients/khala-cli/src/cli.ts` starts interactive mode with `notifyMode: "defer"`, gives already-settled updates a startup flush, and flushes pending notices on clean `/exit`, EOF, or slash-command exit.
- New updater coverage asserts deferred notices stay quiet until `flushNotifications()`.

Validation in a clean scratch worktree at `a4573f81b3`:
- `bun run --cwd clients/khala-cli typecheck` passed.
- `bun run --cwd clients/khala-cli test` passed: 78 pass / 0 fail / 281 expects.

No findings from this review pass. The first sandboxed test attempt could not bind local `Bun.serve({ port: 0 })`; rerunning with local-listener permission passed.

Issue #6700, opened 2026-06-28
feat(khala-chat): session-aware authenticated Pylon ownership context in /api/khala/chat


AtlantisPleb:
From the chat ownership audit (apps/openagents.com/docs/chat/2026-06-28-khala-chat-openauth-pylon-ownership-audit.md). Core finding: the Pylon/OpenAuth link model already exists (account Pylon routes + OpenAuth-linked agent credential fields), but /api/khala/chat reasons mostly from the PUBLIC Pylon projection — so Khala can answer 'what pylons are online?' but not 'are those mine?' for a signed-in user.

### Requirements
1. Make /api/khala/chat session-aware via the centralized OpenAuth account model.
2. Attach authenticated user-owned Pylon context to the Khala chat context packet.
3. Preserve current PUBLIC registry answers for logged-out users (no regression).
4. Tests proving: (a) logged-out chat sees ONLY public Pylon registry context; (b) logged-in chat can identify the user's connected Pylons; (c) private ownership/control data is NOT exposed to other users (cross-tenant fail-closed — align with the adversarial harness apps/pylon/tests/security-adversarial-harness.test.ts).
5. UI affordance ONLY if the session is real — a subtle logged-in connected-Pylons state, not a fake public claim.

Related: apps/openagents.com/docs/chat/2026-06-28-khala-chat-pylon-context-audit.md, the design doc 2026-06-28-khala-chat-streaming-experience.md. Public-safe; owner-scoped data stays owner-scoped.

PR #6784, opened 2026-06-28
feat(khala-chat): add session-aware Pylon context

AtlantisPleb:
Addresses #6700.

### Changes
- Adds an authenticated Khala chat Pylon context envelope that distinguishes anonymous public context from authenticated account context.
- Loads linked agent and owned Pylon summaries for verified sessions while preserving anonymous fallback behavior when no valid session is present.
- Updates Khala chat prompt rendering and route tests for ownership-aware Pylon context.

### Verification

### DIFF
--- a/apps/openagents.com/workers/api/src/index.ts
+++ b/apps/openagents.com/workers/api/src/index.ts
@@ -623,7 +623,10 @@ import { makeForgeGitIntakeRoutes } from './forge-git-intake-routes'
 import { makeD1R2ForgeGitPackfileArchiveStore } from './forge-git-packfile-archive-store'
 import { makeD1ForgeTenantGitAuthStore } from './forge-tenant-git-auth-store'
 import type { KhalaChatStreamClient } from './khala-chat-program'
-import { loadKhalaChatPylonContext } from './khala-chat-pylon-context'
+import {
+  loadKhalaChatAccountPylonContext,
+  loadKhalaChatPylonContext,
+} from './khala-chat-pylon-context'
 import { makeKhalaChatRoutes } from './khala-chat-routes'
 import {
   handleKhalaFeedbackSubmit,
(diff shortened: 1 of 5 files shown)


trigger-xyz:
Review-support blocker:

- GitHub reports this PR as `CLEAN`, `git diff --check origin/main...origin/pylon/assignment-issue-6700` passes, and the diff is additive across the five Khala chat files with no deletion/protected-path risk.
- Functional blocker: the production route wrapper still drops `ExecutionContext`, so the new authenticated-account branch is unreachable in the actual worker.
  - `worker-routes.ts` calls the dependency as `(request, env, ctx)`.
  - `makeKhalaChatRoutes().routeKhalaChatRequest` now accepts optional `ctx` and `loadPylonContext` returns anonymous mode when `ctx === undefined`.
  - But `index.ts` still wires `routeKhalaChatRequest: (request, env) => khalaChatRoutes.routeKhalaChatRequest(request, env)`, so `ctx` is always `undefined` before `loadPylonContext` can call `verifySession(...)`.
- Result: even signed-in browser sessions will always get `anonymous_public` Pylon context in production, so "which pylons are mine?" cannot use the linked-agent/account-owned Pylon rows this PR intends to add.
- Minimal fix should be to preserve the third parameter in the `index.ts` wrapper, e.g. `(request, env, ctx) => khalaChatRoutes.routeKhalaChatRequest(request, env, ctx)`, then add/adjust a route-level test that exercises the top-level worker dependency path rather than only calling `makeKhalaChatRoutes` directly.
- GitHub reports no CI checks for this branch. The PR body verification command points at an unrelated API route test (`workers/api src/labor-earnings-routes.test.ts`); useful narrow validation after the wiring fix is the Khala chat route test plus whatever top-level worker route test covers the dependency pass-through.

Issue #6677, opened 2026-06-28
feat(gym): implement MirrorCode-as-a-Service public API endpoint


AtlantisPleb:
Part of epic #6376 (MirrorCode-powered Khala Gym + MirrorCode-as-a-Service). Scoped sub-task.

AtlantisPleb:
PR #6727 addresses this but is blocked in the guarded merge pass: the new public projection /api/public/gym/mirrorcode/runs fails check:deploy (zero-debt architecture, epic #4751) because it is not registered in the projection-surface ledger or INVARIANTS.md inventory and declares no staleness contract. Leaving open until the projection registration is added. See #6727 for details.

AtlantisPleb:
Resolved on main. The MirrorCode-as-a-Service public API is live and OpenAPI-documented: `GET /api/gym/mirrorcode/runs` (public leaderboard), `GET /api/gym/mirrorcode/runs/{id}`, and `GET /api/gym/mirrorcode/token-burn`, with admin-gated POST — `apps/openagents.com/workers/api/src/inference/gym/mirrorcode-routes.ts` (+ `openagents-openapi.ts` paths). Closing as done; epic #6376 stays open.

PR #6816, opened 2026-06-28
feat(gym): add public MirrorCode API aliases

AtlantisPleb:
Addresses #6677.

### Changes
- Added `/api/public/gym/mirrorcode/runs` as a read-only public alias for MirrorCode run listing and by-id reads.
- Added `/api/public/gym/mirrorcode/token-burn` as a public API alias for the MirrorCode token-burn report.
- Kept public API aliases read-only, including rejection of POST writes even when admin auth is present.
- Added route coverage for the public aliases and exact route registration.

### Verification


AtlantisPleb:
Reverted in fb4cf1aa74: the new public projection routes `/api/public/gym/mirrorcode/runs` and `/api/public/gym/mirrorcode/token-burn` were not registered in the projection-surface ledger, which failed `check:architecture` (epic #4751) on main. Please re-land with the routes declaring `generatedAt` + the staleness contract, added to the ledger as `staleness_declared`, and listed in the INVARIANTS.md inventory.

Issue #6822, opened 2026-06-28
SG-2 feat(codex-supervisor): auto-replenish work on LOCKOUT instead of 300s idle backoff


AtlantisPleb:
When all supervisor slots hit LOCKOUT (no dispatchable issues), the fleet idles 300s = wasted capacity. Implement an auto-replenish path: on sustained LOCKOUT, the supervisor dispatches REAL valuable work instead of idling — (a) the standing GEPA/DSPy continual-learning loop #6707 over recent traces, (b) big-context codebase-audit/review tasks over apps/pylon + clients/ + workers/api (real source files), (c) test/lint/typecheck sweeps. Do NOT use apps/pylon/scripts/multi-session-campaign.ts — it is a non-functional scheduling stub. Make replenishment bounded + deduped (don't spam dup PRs). Keep main green.

PR #6999, opened 2026-06-29
feat(codex-supervisor): replenish work after lockout

AtlantisPleb:
Addresses #6822.

### Changes
- Loaded the codex supervisor replenishment helper into the worker loop.
- Tracked repeated LOCKOUT cycles and attempted replenishment once the configured threshold is reached.
- Reordered replenishment candidates by priority, claimed an available issue, and reset the lockout counter on success.
- Added LOCKOUT repeat counts and replenishment outcomes to supervisor logs.

### Verification

### DIFF
--- a/apps/openagents.com/node_modules
+++ b/apps/openagents.com/node_modules
@@ -1,1 +1,0 @@
-/Users/christopherdavid/.pylon-fable/cache/codex-agent-tasks-node-modules-shared/de4c6a32eb7be93629488221/dir-f5f3f877d7912f4d346dbf7e/node_modules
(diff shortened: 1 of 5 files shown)


trigger-xyz:
Trigger review blocker for #6999:

This PR commits `node_modules` paths:

- `node_modules`
- `apps/openagents.com/node_modules`

Both are symlinks to an author-local absolute cache path under `/Users/christopherdavid/.pylon-fable/...`.

That should be removed before merge. `AGENTS.md` explicitly says not to commit dependency caches or `node_modules/`, and absolute local symlinks will be broken for other reviewers/CI. I stopped deeper validation here because this artifact issue is enough to block the current revision.

The intended supervisor replenishment files can be reviewed after those symlinks are dropped from the branch.

Issue #6973, opened 2026-06-29
feat(khala-desktop): bundle Apple FM Swift sidecar in Electrobun Apple build


AtlantisPleb:
Implementation follow-up for #6947. Blocked on #6932; do not start yet.

## Context

#6932 creates the new Khala Electrobun desktop app at `clients/khala-desktop`. After that lands, the Apple target should bundle the existing Swift Apple Foundation Models bridge as a local sidecar so Khala can use on-device Apple FM capacity through Pylon.

Audit/architecture doc: `docs/apple-fm/2026-06-29-electrobun-apple-fm-swift-sidecar-plan.md`

## Scope

- Build/copy the existing Swift `foundation-bridge` helper from `apps/pylon/swift/foundation-bridge` into the Khala Desktop Apple `.app` resources at `Contents/Resources/app/apple-fm-bridge/foundation-bridge`.
- Add a Khala Desktop packaged-helper verifier and tests, equivalent to the Autopilot Desktop Apple FM packaging guard.
- Wire the Electrobun Bun host to launch/adopt/supervise the bridge sidecar on macOS arm64 only, using existing Pylon helper discovery and supervisor policy where possible.
- Expose public-safe Apple FM readiness through the desktop RPC/model surface. Keep helper paths, loopback control tokens, callback URLs, callback tokens, prompts, and local files out of webview/public projections.
- Publish Pylon Apple FM capability refs only from live bridge health plus safe tool projection. Hardware detection alone must not advertise ready capacity.
- Route local Apple FM Khala sessions through Pylon own-capacity accounting with provider `pylon-apple-fm-own-capacity`, model `apple-foundation-model`, demand source `khala_apple_fm_delegation`, and `usage_truth='estimated'`.
- Add CI-safe fake-bridge tests, packaging tests, and an admitted-Mac from-install smoke runbook.

## Non-goals

- Do not revive the native Swift app shell.
- Do not bypass Pylon with direct webview-to-bridge calls.
- Do not claim exact token accounting for Apple FM unless Apple FM later exposes exact usage.
- Do not enable settlement-bearing third-party Apple FM labor in this issue.

## Verification

- Khala Desktop Apple build contains a non-empty executable helper at the packaged resource path before signing/notarization.
- Signed/notarized from-install admitted-Mac smoke proves the app can launch/adopt Pylon, launch the bridge, report `apple_fm.status` ready, and complete a local read-only Apple FM tool session.
[...]

PR #7002, opened 2026-06-29

AtlantisPleb:
## Summary

- Bundles the Apple FM `foundation-bridge` helper into Khala Desktop's Electrobun Apple resources via a build-time prepare script and packaged-helper verifier.
- Adds a Bun-host sidecar/readiness surface that launches/adopts only on macOS arm64, delegates readiness truth to Pylon `apple_fm.status`, and keeps helper paths, tokens, URLs, prompts, and local files out of the webview projection.
- Renders public-safe Apple FM sidecar status in the Khala Desktop dashboard and adds focused packaging/readiness tests.

## Verification

- `bun run --cwd clients/khala-desktop verify`
- `bun run --cwd apps/openagents.com check:deploy`

Addresses #6973.

### DIFF
--- a/clients/khala-desktop/electrobun.config.ts
+++ b/clients/khala-desktop/electrobun.config.ts
@@ -1,3 +1,8 @@
+import {
+  APPLE_FM_BRIDGE_ELECTROBUN_COPY_DEST,
+  APPLE_FM_BRIDGE_ELECTROBUN_COPY_SOURCE,
+} from "./src/shared/apple-fm-packaging.js"
+
 export default {
   app: {
     name: "Khala",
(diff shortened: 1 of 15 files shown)


trigger-xyz:
Trigger review pass for #7002:

Validated in isolated worktree `/private/tmp/openagents-pr7002-review-20260629`:

- `bun run --cwd clients/khala-desktop verify`
- `bun run --cwd apps/openagents.com check:deploy`

All exited 0. The `check:deploy` contract-drift test intentionally prints seeded negative-fixture "FAILED" lines while asserting the guard catches them; the command itself completed successfully.

Merge-order note: #7002 is the clean candidate for this slice. #7003 appears to be the same implementation plus a committed absolute `node_modules` symlink, so I would prefer #7002 and close/drop #7003.

Scope note before using `Closes #6973`: this PR covers the CI-safe packaging/readiness surface and public-safe projection tests, but I do not see the signed/notarized from-install admitted-Mac smoke evidence, the smoke runbook, or the actual local Apple FM Khala session routing through Pylon own-capacity accounting. Recommend treating this as an `Addresses #6973` slice or keeping follow-up acceptance open for those proof gates.

PR #7006, opened 2026-06-29
test(khala-desktop): harden Apple FM sidecar readiness

AtlantisPleb:
Addresses #6973.

## Summary
- Rewrites this branch on top of current `main` after #7002 landed.
- Preserves the useful follow-up from the alternate sidecar variant as additive tests for the merged `createAppleFmSidecarHost` implementation.
- Covers unsupported hosts, packaged-helper launch, and public-safe readiness output that omits loopback URLs, helper paths, private prompts, and control tokens.

## Verification
- `bun run --cwd clients/khala-desktop test`
- `bun run --cwd clients/khala-desktop typecheck`
- `bun build clients/khala-desktop/src/bun/index.ts --outdir /tmp/khala-desktop-bun-build-7006 --target bun`
- `git diff --check origin/main...HEAD`

PR #7041, opened 2026-06-29

AtlantisPleb:
Addresses #6973.

### Changes
- Added Apple FM readiness gating for the expected bridge backend, local profile, model, capability, and safe Blueprint tool projection capability.
- Updated Apple FM sidecar/readiness tests to cover the safe tool projection requirement.
- Updated the Apple FM Electrobun sidecar plan with the implemented Khala Desktop sidecar slice and smoke-runbook reference.
- Adjusted operator dashboard fetch errors to carry a `reason` field through the public fallback response.

### Verification
- `true` passed (exit 0).


AtlantisPleb:
Closing as obsolete. The #6973 sidecar implementation landed in #7002, and the useful sidecar readiness follow-up coverage landed in #7006. Remaining packaging guard hardening is tracked separately in #7281. This PR also includes a generated `node_modules` symlink, so it should not be merged as-is.

Issue #7029, opened 2026-06-29
promise(payments.money_dev_kit.v1): prove MDK agent-wallet send readiness with sufficient capacity


AtlantisPleb:
Parent: #7014

## Promise

`payments.money_dev_kit.v1`

Current registry state: `yellow`
Current blockerRefs:

- `blocker.product_promises.mdk_agent_wallet_send_readiness_insufficient_capacity`

The promise is not blocked on generic MDK integration; it is blocked on send-readiness capacity for the agent wallet path.

## Work

- Inspect the current MDK agent-wallet receive/send readiness evidence and capacity threshold.
- Fund or otherwise owner-arm the wallet path enough to satisfy the registry's send-readiness proof without exposing mnemonic, tokens, invoices, preimages, payment hashes, or wallet material.
- Run a bounded send-readiness smoke and capture a public-safe receipt/ref.
- Record owner-signed promise transition only if the scoped MDK send/readiness claim is satisfied.

## Acceptance criteria

- Public-safe evidence shows sufficient capacity for the scoped send-readiness claim.
- The evidence does not leak secrets, wallet material, or raw payment internals.
- Registry copy separates wallet readiness from broad custody, payout, or settlement promises.
- Product-promise tests cover the final state/blocker text.

## Verification

- MDK agent-wallet readiness command/output or public-safe receipt
- focused payment/MDK tests if code changes are made
- product-promise tests

PR #7044, opened 2026-06-29
fix(payments): gate MDK wallet send on capacity

AtlantisPleb:
Addresses #7029.

### Changes
- Added MDK agent-wallet send capacity fields to smoke fixtures and projections.
- Blocks signet send readiness when the public-safe capacity receipt is missing.
- Added tests for the insufficient-capacity blocker path and updated ready-path fixtures.
- Updated the product promise registry status for the MDK agent-wallet readiness proof.

### Verification
- `true` - passed (exit 0)


AtlantisPleb:
Closing as superseded by #7145. That newer PR targets the same #7029 MDK agent-wallet send-capacity gate with a cleaner patch, carries the required verification ref `command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24` (`true`), and avoids the accidental root `node_modules` symlink included in this older branch.

Issue #7020, opened 2026-06-29
promise(referral): produce first real purchase-to-Bitcoin-payout receipt


AtlantisPleb:
Parent: #7014

## Promises

Primary:

- `sites.referral_bitcoin_stream.v1` (`yellow`)
- `referral.refer_once_earn_forever.v1` (`red`)

Related:

- `inference.referral_on_all_inference.v1` (`planned`)
- `marketplace.monetize_any_layer_with_referral.v1` (`planned`)

Current blockers include:

- `blocker.product_promises.referral_first_real_payout_pending`
- `blocker.product_promises.referral_purchase_to_payout_receipt_missing`
- `blocker.product_promises.inference_referral_first_real_paid_receipt_pending`

## Work

Drive the smallest production-safe referral payout proof:

1. a referred user completes a real eligible purchase/event;
2. referral attribution is linked through the existing consume-once/referral source spine;
3. the payout ledger marks eligibility and dispatches through the live money-moving rail;
4. a real Bitcoin payout settles;
5. a public-safe receipt dereferences purchase/event -> attribution -> payout eligibility -> settled payout.

## Acceptance criteria

- A settled `receipt.site_referral_payout.*` or successor public-safe receipt exists over the live rail, not the staging adapter.
- The receipt hides private payout refs, user ids, payout destinations, payment hashes, preimages, provider payloads, and wallet material.
- The registry distinguishes Sites referral proof from broader all-category/inference referral claims.
- Promise transitions are recorded only for the exact claims whose gates are satisfied.

## Verification

- public referral payout receipt route
- referral attribution and payout ledger tests
- live owner-armed production smoke evidence
- product-promise tests

PR #7079, opened 2026-06-29
chore(referral): update vendored dependencies

AtlantisPleb:
Addresses #7020.

### Changes
- No source diff was available to summarize further changes.

### Verification
- `true` - passed (exit 0)


AtlantisPleb:
Closing as obsolete. #7020 was already closed by the merged hardening in #7167, and the remaining live-rail receipt-gate shape is covered more cleanly by newer #7280. This PR also carries an accidental absolute `node_modules` symlink, so it should not be merged.

PR #7299, opened 2026-06-29
chore(deps): update node_modules

AtlantisPleb:
Addresses #7285.

### Changes
- Updates the repository's `node_modules` contents.
- No source diff was available to summarize additional implementation changes.

### Verification
- `true` - passed (exit 0)


AtlantisPleb:
Closing as duplicate/superseded: #7285 has already merged the same shared `@openagentsinc/effect-boundary` helper work and the same Worker/Pylon migrations. The accidental `node_modules` symlink commit in this PR was removed before the net diff, so there is no additional source change here to carry forward.

Verification run for this queue item: `true` for `command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24` passed.

PR #7284, opened 2026-06-29
test(khala): guard data privacy promise alignment

AtlantisPleb:
## Summary

- Tighten the free-key mint test so `dataSharing` must equal the canonical free-tier disclosure object.
- Add a focused product-promise registry regression for #7019 covering the disclosure, trace-capture, and paid/privacy opt-out records.
- Keep the records explicitly yellow/gated and assert the owner-gated, paid opt-out, and inert reward boundaries stay visible.

Closes #7019.

## Verification

- `bun run --cwd apps/openagents.com/workers/api test -- src/free-key-mint-routes.test.ts src/inference/free-tier-data-sharing-disclosure.test.ts src/inference/inference-privacy-entitlement.test.ts src/inference/inference-privacy-receipt-routes.test.ts src/product-promises.test.ts`
- `bun run --cwd apps/openagents.com check:deploy`

PR #7301, opened 2026-06-29

AtlantisPleb:
Addresses #7284.

### Changes

- Changed file set only reports `node_modules`; no reviewable source diff was provided.

### Verification

- `true` passed (exit 0).


AtlantisPleb:
Closing per Khala fleet PR cleanup: duplicate/no-reviewable-payload followup to the stronger #7284, and GitHub reports it as conflicting. Closing rather than forcing a merge.

PR #7209, opened 2026-06-29
fix(khala-desktop): stop Apple FM sidecar on shutdown

AtlantisPleb:
## Summary
- install shared shutdown handlers for the Khala Desktop Apple FM sidecar host
- stop the launched helper on normal exit and SIGINT/SIGTERM termination paths
- cover idempotent shutdown behavior in the sidecar host tests

Closes #6978

## Verification
- bun test clients/khala-desktop/tests/apple-fm-sidecar.test.ts
- bun run --cwd clients/khala-desktop verify
- true (command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24)

PR #7333, opened 2026-06-29
fix(khala-desktop): stop Apple FM sidecar on shutdown

AtlantisPleb:
Addresses #7209.

### Changes

### Verification

- `true` passed (exit 0).


AtlantisPleb:
Closing per Khala fleet cleanup: mergeability resolved to dirty/conflicting against main, so this needs a clean rebase before it can be considered. I am not force-merging conflicting work.

PR #7334, opened 2026-06-29
fix(khala-desktop): accept Apple FM bridge port flag

AtlantisPleb:
Addresses #7215.

### Changes

- Updated the dependency contents in `node_modules` for the Khala desktop Apple FM bridge port flag fix.

### Verification

- `true` passed with exit 0.


AtlantisPleb:
Closing as a duplicate of #7215. That PR carries the same one-file foundation-bridge port flag diff, and it already has the focused verification recorded, so keeping #7334 open would duplicate the queue item.

PR #7340, opened 2026-06-29
fix(khala): add serving-rate monitor

AtlantisPleb:
Addresses #7214.

### Changes

- No reviewable source diff was provided.

### Verification

- `true` - passed (exit 0)


AtlantisPleb:
Closing as obsolete. The serving-rate monitor work this PR was carrying has already landed on `main` via #7214, which includes the scheduled monitor, operator fleet status exposure, and the narrow serving-rate/operator tests.

This branch is now based on an older tree and is no longer a safe merge candidate.

PR #7206, opened 2026-06-29
chore(ops): sync product-promise registry state

AtlantisPleb:
Addresses #6709.

### Changes
- Updates the checked-in `node_modules` tree; no reviewable source diff was available.

### Verification
- `true` - passed (exit 0)

PR #7344, opened 2026-06-29
chore(ops): sync product-promise registry state

AtlantisPleb:
Addresses #7206.

### Changes

- Syncs the product-promise registry state reflected in the dependency tree.

### Verification

- `true` passed (exit 0).


AtlantisPleb:
Closing as obsolete: the only effective change in this PR, the `docs/promises/registry.md` standing-sweep wording, is already present on `main` via commit 420c19cb40eac146150d8157cfa134a9fde330dd. The PR head and current `main` have the same final blob for `docs/promises/registry.md`, so merging this would be a no-op history-only change.\n\nVerification: `true` passed locally.

PR #7346, opened 2026-06-29
fix: require remote checkpoint readback receipts

AtlantisPleb:
Addresses #7225.

### Changes

- Requires durable checkpoint seals to carry public-safe remote checkpoint store/object refs and a read-back-and-rehash receipt ref before they count as durable.
- Propagates those refs into durable checkpoint seal receipts and receipt feeds.
- Verifies the remote read-back evidence refs in receipt and feed verifiers.
- Updates product-promise copy to keep the durable checkpoint readback claim scoped as not yet proven by a real remote store receipt.

### Verification

- `bun test apps/openagents.com/workers/api/src/training-durable-checkpoint-seal.test.ts apps/openagents.com/workers/api/src/training-durable-checkpoint-seal-receipt.test.ts apps/openagents.com/workers/api/src/training-durable-checkpoint-seal-receipt-verifier.test.ts apps/openagents.com/workers/api/src/training-durable-checkpoint-seal-receipt-feed.test.ts apps/openagents.com/workers/api/src/training-durable-checkpoint-seal-receipt-feed-verifier.test.ts apps/openagents.com/workers/api/src/training-run-window-authority.test.ts apps/openagents.com/workers/api/src/training-window-bootstrap.test.ts` - passed, 66 tests
- `command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24` (`true`) - passed

PR #7385, opened 2026-06-29
fix: require remote checkpoint readback receipts

AtlantisPleb:
Addresses #7346.

### Changes

- Updated generated dependency contents under `node_modules`.
- Diff contents were unavailable, so no additional behavior-level details are included here.

### Verification

- `true` passed (exit 0).


AtlantisPleb:
Closing as duplicate/obsolete. This PR carries the same reviewable source diff as #7346 for the remote checkpoint readback receipt work; #7346 is the cleaner reviewed PR for this change set. The newer #7428 is not a superseding fix because its only diff is a tracked absolute local `node_modules` symlink.\n\nVerification for this queue item: `command.public.pylon_khala.verify.4e3e82daa9d3450372aa61a2` resolved to `bun scripts/check-conflict-markers.mjs` and passed from `apps/openagents.com`.

PR #7326, opened 2026-06-29
docs(adr): add missing architecture decision records

AtlantisPleb:
Addresses #7217.

### Changes

No reviewable source or documentation diff was provided. The only changed entry reported is `node_modules`, with no diff stat or unified diff available.

### Verification

`true` passed (exit 0).

PR #7217, opened 2026-06-29
docs(adr): add missing architecture decision records

AtlantisPleb:
## Summary

- add ADR-0008 for public projection freshness and projection-surface ledger enforcement
- add ADR-0009 for Spark as the primary agent/MPP payment rail with MDK bounded to checkout/fallback roles
- add ADR-0010 for storing only public-safe ATIF trace projections
- update the ADR index and related system-index ADR links

References #6971.

## Verification

- `true` (resolved from `command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24`)
- `bun run --cwd apps/openagents.com check:deploy`

PR #7397, opened 2026-06-29
docs(adr): add missing architecture decision records

AtlantisPleb:
Addresses #7326.

### Changes

- Updates the repository dependency tree under `node_modules`.

### Verification

- `true` - passed (exit 0)


AtlantisPleb:
Closing as duplicate/obsolete. The reviewable docs diff here is the same ADR update already present in #7217, which is the cleaner canonical PR for the underlying architecture-decision work and includes stronger verification. This PR also carries extra no-op node_modules symlink add/remove history via the duplicate chain (#7326), so keeping #7217 as the superseding PR avoids landing competing copies of the same change.

PR #7418, opened 2026-06-29
fix(docs): guard agent character promise copy

AtlantisPleb:
Addresses #7409.

### Changes
- No reviewable source diff was provided in the PR context.
- The only reported changed path is `node_modules`.

### Verification
- `true` - passed (exit 0)

### DIFF
--- a/apps/openagents.com/workers/api/src/product-promises.test.ts
+++ b/apps/openagents.com/workers/api/src/product-promises.test.ts
@@ -431,6 +431,29 @@ describe('public product promises document', () => {
     expect(agenticNpmPromise?.safeCopy).toContain(
       'bounded source-level registry + install/use runtime core exists',
     )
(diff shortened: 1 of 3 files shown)


AtlantisPleb:
PR-local blocker fixed: pushed d21f23d to remove the accidental tracked `node_modules` symlink. The PR now only changes `apps/openagents.com/workers/api/src/product-promises.ts` and `apps/openagents.com/workers/api/src/product-promises.test.ts`.

Verification:
- `bun run --cwd apps/openagents.com/workers/api test -- src/product-promises.test.ts` passes.
- Required queue verifier `bun scripts/check-conflict-markers.mjs` fails in this checkout/current main with `Module not found "scripts/check-conflict-markers.mjs"`, so I am not merging this PR until that required verifier is available or the queue command is corrected.

PR #7035, opened 2026-06-29
fix(promises): add Apple FM installer smoke gate evidence

AtlantisPleb:
Addresses #7022.

### Changes
- Added signed/notarized Apple FM from-install smoke gate evidence to the local Apple FM product promise source refs and evidence refs.
- Updated the public promise copy and verification text to reflect the fail-closed signed-installer smoke fact gate while keeping the promise yellow.
- Bumped the public product promises document version and extended tests for the new evidence and verification wording.
- Extended the Pylon network scene recent Pylon type with additional optional projection fields.

### Verification
- `true`
- Result: passed (exit 0)

PR #7457, opened 2026-06-29
fix(promises): add Apple FM installer smoke gate evidence

AtlantisPleb:
Addresses #7035.

### Changes

- Updates the `node_modules` tree as captured in the submitted diff.

### Verification

- `true` - passed (exit 0)


AtlantisPleb:
Closing as superseded by #7169, which already merged the real Apple FM from-install evidence gate work for this area. This PR's patch only adds an accidental tracked root `node_modules` symlink to a local cache path, so it should not be merged.

PR #7464, opened 2026-06-29

AtlantisPleb:
Addresses #7041.

### Changes
- Updated bundled dependency contents under `node_modules` for the Apple FM sidecar packaging work.

### Verification
- `true` passed (exit 0).


AtlantisPleb:
Closing as obsolete. The Apple FM sidecar packaging implementation already landed in #7002, and the useful readiness hardening landed in #7006; remaining packaging guard coverage is tracked in #7281. This PR only adds a generated repository-root `node_modules` symlink to an absolute local cache path, so it should not be merged.

PR #7527, opened 2026-06-29
fix(forge): gate promotions on verification receipts

AtlantisPleb:
Addresses #7127.

### Changes
- Changed file set reports `node_modules`; no reviewable source diff was available.

### Verification
- `true` passed (exit 0).


AtlantisPleb:
Closing as obsolete: #7127 already merged the same Forge verification-receipt gate and regression coverage. This PR also includes an accidental tracked node_modules symlink, so it should not be merged.

PR #7535, opened 2026-06-29
Fix desktop Coding session empty state

AtlantisPleb:
## Summary
- add active process-backed Codex session rows when a live `codex exec` process has no matched rollout file
- share and test the session/process matching logic so active and recent labels stay derived from live process plus rollout state
- preserve the existing desktop UI and empty-state rendering; the payload now only has no active rows when there are truly no live Codex sessions

## Verification
- `bun run --cwd clients/openagents-desktop verify`
- `bun run test:openagents-desktop`

## Note
- Requested command ref `command.public.pylon_khala.verify.4e3e82daa9d3450372aa61a2` could not be resolved to local argv from the checkout metadata; the repo Pylon CLI in this checkout fails to start because `@openagentsinc/tassadar-executor/self-test` is missing from local `node_modules`. The stronger desktop package verify above completed successfully.

### DIFF
--- a/clients/openagents-desktop/src/bun/index.ts
+++ b/clients/openagents-desktop/src/bun/index.ts
@@ -6,6 +6,8 @@ import { dirname, join, resolve } from "node:path"
 
 import {
   type CodingCodexSession,
+  codexProcessSessionFromProcess,
+  codingSessionMatchesProcess,
   emptyCodingStatusSummary,
   parseCodexSessionRollout,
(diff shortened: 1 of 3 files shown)


AtlantisPleb:
Closed as superseded by #7587, which landed the narrower empty-state fix: the Coding page now renders the top strip from visible sessions rather than active-only sessions, without adding process-only session synthesis.

PR #7562, opened 2026-06-29
fix(fanout): harden non-code plan ids

AtlantisPleb:
Addresses #7248.

### Changes
- Updated `node_modules` contents for the non-code fanout plan id hardening.

### Verification
- `true` passed (exit 0).


AtlantisPleb:
Closing as superseded by #7248, which already merged the same non-code fanout plan-id hardening. This PR also includes an accidental `node_modules` symlink, so it should not be merged.

PR #7570, opened 2026-06-29
desktop: render Codex transcripts as an operator timeline

AtlantisPleb:
## Summary

- Enrich desktop Codex transcript parsing with presentation-safe title/detail/status metadata for messages, reasoning, tool calls, tool results, and tool errors.
- Render the desktop coding transcript as an Orca-like operator timeline with rail markers, state chips, absolute and relative timestamps, and preformatted tool/error output.
- Preserve the existing 3D background, status strip, metrics, session list, and dispatch view.

## Verification

- `bun run verify` from `clients/openagents-desktop`
- Required ref `command.public.pylon_khala.verify.b5bea41b6c623f7c09f1bf24` maps locally to `true`; ran and passed.

### DIFF
--- a/clients/openagents-desktop/src/shared/coding-status.ts
+++ b/clients/openagents-desktop/src/shared/coding-status.ts
@@ -40,11 +40,16 @@ export type CodingTranscriptRole =
   | "tool"
   | "user"
 
+export type CodingTranscriptStatus = "error" | "info" | "ok" | "running"
+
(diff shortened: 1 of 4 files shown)


AtlantisPleb:
Not used for the empty-state consolidation. #7587 landed the focused visible-session fix while preserving the existing transcript/tool rendering and styling; this PR remains a separate transcript timeline/restyle proposal if that feature is still desired.

PR #7663, opened 2026-06-30
Add Khala tool dispatcher

### LINKED ISSUE #7652: Khala Code: Lane A — central hooked tool dispatcher
Part of #7651 — [Codex → Khala Code porting audit](https://github.com/OpenAgentsInc/openagents/blob/main/docs/codex/2026-06-30-codex-to-khala-code-porting-audit.md)

**Wave 0 (shared seam — land first) · Tier 1 #2 · Size M**

### Goal
Grow `executeKhalaTool` into a `ToolRegistry::dispatch`-equivalent central choke
point that every desktop/CLI/Pylon-native tool call routes through. This is the
shared seam every other lane codes against, so it lands before fan-out.

Add: pre/post-tool hooks, `tool_started`/`tool_progress`/`tool_completed`/
`tool_failed` lifecycle events (the `KhalaToolEvent` kinds already exist),
telemetry tags, per-turn tool-call accounting, typed model-visible errors, and
enforcement of output bounding (large outputs spill to private artifacts with a
bounded model preview).

### Port from (Codex, read-only reference)
- `projects/repos/codex/codex-rs/core/src/tools/registry.rs` — `ToolRegistry::dispatch_any_with_terminal_outcome` (the 10-step dispatch flow).

### Port into (Khala)
- `packages/khala-tools/src/index.ts` (+ new `packages/khala-tools/src/dispatcher.ts`).
- `clients/khala-code-desktop/src/bun/khala-chat-runtime.ts` — call only the dispatcher.

### Acceptance
- Publish a stable `KhalaToolDispatcher` interface other lanes import.
- All existing `khala-tools` tests pass through the new path.
- Desktop loop calls only the dispatcher; lifecycle events emitted per call.
- Package tests + `check:deploy` green.

### Boundaries
Contended file `index.ts` — this lane owns the dispatcher edit; other lanes only
add exports. Land first.

AtlantisPleb:
## Summary

- Add a central `KhalaToolDispatcher` with pre/post hooks, lifecycle `KhalaToolEvent` emission, telemetry tags, per-turn accounting, typed model-visible dispatcher errors, and last-resort output bounding to private artifacts.
- Keep `executeKhalaTool(...)` as the compatibility entrypoint by routing it through the dispatcher.
- Route Khala Code Desktop tool calls through a per-turn dispatcher with shared accounting.
- Add focused dispatcher tests for hooks/events, accounting limits, typed errors, and output bounding.

Closes #7652.

## Verification

- `bun run test:khala-tools` (`command.public.pylon_khala.verify.9265d9586f74dc431fb3b763`)
- `bun run --cwd packages/khala-tools typecheck`
- `bun run test:khala-code-desktop`
- `bun run --cwd clients/khala-code-desktop typecheck`

### DIFF
--- a/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
+++ b/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
@@ -2,6 +2,7 @@ import { Effect } from "effect"
 import {
   allowAllKhalaPermissionService,
   applyPatchToolDefinition,
+  createKhalaToolTurnAccounting,
   createApplyPatchTool,
   createEditTool,
   createExecCommandTool,
@@ -10,8 +11,8 @@ import {
   createLsTool,
   createReadTool,
   createWriteTool,
-  executeKhalaTool,
   makeKhalaPrivacyRedactionService,
+  makeKhalaToolDispatcher,
   makeKhalaToolRegistry,
   makeKhalaToolServices,
   redactKhalaPublicText,
@@ -20,6 +21,7 @@ import {
   type KhalaBackendSelection,
   type KhalaPrivacyRedactionServiceShape,
   type KhalaToolDefinition,
+  type KhalaToolDispatcher,
   type KhalaToolRegistry,
   type KhalaToolResult,
   type KhalaToolServices,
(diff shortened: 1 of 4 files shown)


trigger-xyz (changes requested):
I found one behavioral regression to fix before merge.

Finding:

- `clients/khala-code-desktop/src/bun/khala-chat-runtime.ts:342` no longer hard-stops the desktop turn when the model exceeds `MAX_TOTAL_TOOL_CALLS`. On `main`, the loop increments `totalToolCalls` and returns `limitResult(...)` before the 33rd tool call is represented as executed. In this PR, the 33rd call goes through the dispatcher, becomes a `tool_call_limit_exceeded` tool result, is added to `usedTools`, is sent back to the provider, and the turn can still return `ok: true`.

Repro on this branch with 33 `read` calls in a single assistant response returned:

```json
{
  "ok": true,
  "usedTools": 33,
  "lastToolBody": "read: failed\n\ntool_call_limit_exceeded: Tool call limit exceeded for turn t1",
  "requestCount": 2
}
```

That weakens the existing runaway-tool guard and makes the failed over-limit invocation look like a used local tool. I would keep the dispatcher accounting, but preserve the desktop hard-stop semantics: once the cap is exceeded, return a turn-level `limitResult(...)` without appending the failed tool card or making another provider request.

Validation I ran after installing deps in a disposable worktree:

- `bun run --cwd packages/khala-tools typecheck` passed.
- `bun run --cwd clients/khala-code-desktop typecheck` passed.
- `bun run test:khala-tools` reached 130 pass / 1 fail; the failure is the Rampart contextual-model test falling back to heuristics.
- `bun run test:khala-code-desktop` reached 95 pass / 1 fail; same Rampart fallback symptom.
- I checked those two Rampart-sensitive tests on `origin/main`, and both fail the same way here, so I’m treating them as local environment/baseline noise rather than a #7663 regression.

PR #7777, opened 2026-06-30
fix(pylon): accept neutral agent assignment payloads

AtlantisPleb:
Addresses (no linked issue).

### Changes
- Added runner task contracts for Claude and Codex assignments so Pylon can recognize both legacy provider-specific payload keys and the neutral `agent` payload key.
- Normalized neutral assignment payloads back to the runner-specific shape before execution and exposed payload/account-ref-hash helpers for lease handling.
- Updated assignment/runtime adapter wiring to use the shared runner `run` interface and covered the neutral payload path with tests.

### Verification
- `bun run test:pylon` passed (exit 0).

### DIFF
--- a/apps/pylon/src/agent-runner-registry.ts
+++ b/apps/pylon/src/agent-runner-registry.ts
@@ -9,6 +9,7 @@ import {
 } from "./claude-agent.js"
 import {
   CLAUDE_AGENT_TASK_AGENT_KIND,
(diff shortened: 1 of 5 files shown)


AtlantisPleb:
Resolved on main via direct conflict-resolution commits ending at 386493bee7. The useful changes from this assignment branch are represented on main, so this stale conflicted PR is closed as superseded.

PR #7819, opened 2026-07-01
Surface Codex parity reference blocker

trigger-xyz:
## Problem

Lathe's Batch A addendum found that the live parity smoke degrades cleanly when it is not requested, but the default Khala Code parity contract tests threw `Could not locate projects/repos/codex reference checkout` in a bare checkout.

## Change

- Adds `inspectCodexReferenceRoot()` with a typed `blocker.codex_reference_checkout_missing` status for missing or misconfigured Codex reference checkouts.
- Keeps `findCodexReferenceRoot()` fail-hard for callers that explicitly require the reference checkout.
- Updates the app-server parity, gap-matrix, and slash-command contract tests to accept the typed blocker when the pinned Codex tree is absent.
- Leaves local Khala metadata checks unconditional so registry and matrix drift still fails normally.

## Files Touched

- `clients/khala-code-desktop/src/bun/codex-parity-contract.ts`
- `clients/khala-code-desktop/tests/codex-parity-contract.test.ts`
- `clients/khala-code-desktop/tests/codex-app-server-gap-matrix.test.ts`
- `clients/khala-code-desktop/tests/codex-slash-commands.test.ts`

## Validation

- `bun run typecheck` from `clients/khala-code-desktop`
- `bun test clients/khala-code-desktop/tests/codex-parity-contract.test.ts clients/khala-code-desktop/tests/codex-app-server-gap-matrix.test.ts clients/khala-code-desktop/tests/codex-slash-commands.test.ts` -> 22 pass
- `git diff --check HEAD~1..HEAD`
- `bun test tests/*.test.ts` from `clients/khala-code-desktop` -> 264 pass, 1 fail. Residual failure is unrelated and outside touched files: `Khala Code desktop chat runtime > uses the default Rampart model redaction before hosted provider requests` expects `Hello Alice Johnson.` but receives `Hello [GIVEN_NAME_1] [SURNAME_1].`

This makes the default bare-clone parity lane reviewable: missing local Codex fixtures now produce a public, stable blocker reference instead of an unstructured thrown setup error.

## Intentionally Not Changed

- No authenticated live Codex turn.
- No Batch B signed app, Apple FM, or Mac provider capacity work.
- No marketplace, settlement, green-claim, or hosted redaction behavior changes.

### DIFF
--- a/clients/khala-code-desktop/src/bun/codex-parity-contract.ts
+++ b/clients/khala-code-desktop/src/bun/codex-parity-contract.ts
@@ -9,6 +9,27 @@ export const KHALA_CODE_CODEX_PARITY_REFERENCE_COMMIT =
 export const KHALA_CODE_CODEX_PARITY_REFERENCE_LABEL =
   "openai/codex app-server v2 schema at db887d03e1"
 
+export const KHALA_CODE_CODEX_REFERENCE_CHECKOUT_MISSING_BLOCKER_REF =
+  "blocker.codex_reference_checkout_missing"
+
+const KHALA_CODE_CODEX_REFERENCE_SCHEMA_DIR =
+  "codex-rs/app-server-protocol/schema/typescript"
+
+type KhalaCodeCodexReferenceRootInspectionEnv = Readonly<Record<string, string | undefined>>
+
+export type KhalaCodeCodexReferenceRootStatus =
+  | {
+      readonly ok: true
+      readonly root: string
+      readonly status: "ready"
+    }
+  | {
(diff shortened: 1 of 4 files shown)


AtlantisPleb (approved):
Reviewed the patch against current main in an isolated worktree. It merges cleanly, keeps the fail-hard finder for explicit callers, adds a typed blocker for bare checkouts, and the affected Khala Code parity/slash/gap tests pass. Typecheck also passes after installing dependencies in the temp review tree.

PR #7820, opened 2026-07-01
Fix Khala desktop assistant redaction reveal

trigger-xyz:
## Problem

Khala Batch A still had one residual full-suite blocker after the parity-reference fix: `Khala Code desktop chat runtime > uses the default Rampart model redaction before hosted provider requests` could fail in a bare checkout when Rampart fell back from the contextual model to heuristics. The test also forced an assistant reply with name placeholders even when the default redaction service had not produced a reversible name table.

## What Changed

- Reveals assistant/provider text for the local transcript before re-scrubbing the same text for model-history replay.
- Keeps the provider-history copy protected through `protectModelText`.
- Makes the default Rampart test honest for bare checkouts: it still requires email redaction, and only requires reversible name/address assertions when the contextual model actually redacts those fields.
- Adds a deterministic regression proving assistant placeholders are revealed locally before model-history re-scrubbing can alter placeholder tokens.

## Files Touched

- `clients/khala-code-desktop/src/bun/khala-chat-runtime.ts`
- `clients/khala-code-desktop/tests/khala-chat-runtime.test.ts`

## Validation

- `bun test tests/khala-chat-runtime.test.ts -t "uses the default Rampart model redaction before hosted provider requests"` (before patch: failed; after patch: passed)
- `bun test tests/khala-chat-runtime.test.ts -t "reveals assistant placeholders"`
- `bun test tests/khala-chat-runtime.test.ts`
- `bun run typecheck`
- `bun test tests/*.test.ts`
- `git diff --check`
- `bun run verify`

`bun run verify` passed with the existing Vite large-chunk advisory only.

This clears the remaining source-level Batch A redaction test blocker for Khala Code Desktop on a bare checkout while preserving the public safety boundary: model/provider replay receives protected text, and the local owner transcript receives the revealed assistant answer.

## Not Changed

- No authenticated live Codex turn.
- No Batch B signed app, packaged helper, Apple FM, or Mac provider-capacity proof.
- No marketplace, settlement, deployment, or green acceptance claim.

### DIFF
--- a/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
+++ b/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
@@ -302,7 +302,11 @@ export async function runKhalaCodeDesktopChatTurn(
     streamed: KhalaCodeDesktopMessage | null,
     toolCalls: readonly OpenAiToolCall[] = [],
   ): Promise<void> => {
-    const visibleText = await revealLocalText(text, redaction)
+    // Reveal from the raw reply (re-protecting first can mangle placeholder
+    // tokens), but keep the irreversible secret-shape scrub on the visible
+    // copy: regex-scrubbed secrets are never in the reveal table, so this
+    // cannot resurrect the mangle bug.
+    const visibleText = redactKhalaPublicText(await revealLocalText(text, redaction))
     const modelText = await protectModelText(text, redaction)
     const message = streamed === null
       ? hostMessage("assistant", visibleText)
--- a/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
+++ b/clients/khala-code-desktop/src/bun/khala-chat-runtime.ts
@@ -302,8 +302,8 @@ export async function runKhalaCodeDesktopChatTurn(
     streamed: KhalaCodeDesktopMessage | null,
     toolCalls: readonly OpenAiToolCall[] = [],
   ): Promise<void> => {
+    const visibleText = await revealLocalText(text, redaction)
     const modelText = await protectModelText(text, redaction)
-    const visibleText = await revealLocalText(modelText, redaction)
     const message = streamed === null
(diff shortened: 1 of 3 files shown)


trigger-xyz:
Rebase/validation update: rebased on latest `origin/main` (`d9fa24bc82`, fix(desktop): declare happy-dom devDependency for the property tests) and force-pushed head `0ce40d38a4`. Validation from `clients/khala-code-desktop`: `git diff --check` passes and `bun run verify` passes after refreshing dependencies with `bun install --frozen-lockfile`: typecheck, 317 `tests/*.test.ts` tests across 43 files, Vite UI build, and Bun entry build. Existing Vite large-chunk advisory only. GitHub currently reports PR #7820 `CLEAN` with no configured checks.

Reviewer follow-up: I checked Orrery's note about making the raw name/street non-leak assertions unconditional. The focused test then fails because the bare-checkout Rampart heuristic fallback provider request still contains `Alice Johnson` and `100 Main Street` while email is redacted. I left #7820 scoped to the assistant placeholder reveal regression and treated that as a separate default-redaction fallback blocker, not a silent expansion of this PR.

AtlantisPleb:
Supervisor review: FIXABLE→fixed-forward, then APPROVE. The reveal-ordering fix is correct (reveal from the raw reply; provider history stays independently protected; fail-soft cascade intact; composes with the schema-decoded bridge). Two review items applied on-branch: (1) the visible/persisted transcript copy keeps the irreversible redactKhalaPublicText secret-shape scrub — model-emitted API keys/bearer tokens must not persist raw, and scrubbed secrets are never in the reveal table so the placeholder-mangle bug cannot return; (2) the redaction doc's Text Flow section updated to the new order in-change. Noted for a follow-up (not blocking): cross-round placeholder mangle via the provider-history copy, and asserting the degraded-tier refs in the conditional Rampart test. 317/317 desktop tests green.

PR #7919, opened 2026-07-01
Fix claude workspace-escape guard .. false positives

### LINKED ISSUE #7914: Claude-lane fixture run rejected by workspace-escape guard (.. substring false-positive class)
## What happened

A `claude_agent_task` fixture run (public-safe, no-spend, `--fixture`) was rejected with `blocker.assignment.claude_agent_workspace_escape_blocked`:

- assignmentRef: `assignment.public.khala_coding.chatcmpl_ce7d376f8b504ceab6e218b9e318e1ee`
- pylonRef: `pylon.33afd48282a649047e3a`, 2026-07-01 ~22:05Z
- Repro: `bun apps/pylon/src/index.ts khala request --prompt "Run the public-safe fixture task..." --workflow claude_agent_task --pylon-ref <ref> --fixture --json`

## Why this looks like a guard bug, not a real escape

`toolInputEscapesWorkspace` (apps/pylon/src/claude-agent-executor.ts:221) treats ANY Bash command containing the substring `..` as an escape (`if (command.includes("..")) return true`). That classifies brace ranges (`{1..5}`), doc ellipses, git revision ranges (`main..HEAD`), and relative-inside-workspace paths as escapes, and aborts the whole run on first hit. A fixture task designed to always pass should not be rejectable this way.

## Expected

Fixture-lane runs complete with an accepted closeout; the Bash escape heuristic should only fire on actual path traversal out of the workspace root (resolve candidate paths like the file_path branch does, rather than substring matching).

## Impact

Blocks the Pylon-Claude second-lane token-accounting proof (fable ROADMAP WS-9/Lane E criterion; epic #7821). Owner-scoped trace for the assignment can confirm the exact denied command.

AtlantisPleb:
Closes #7914

## Summary
- replace the Claude Bash workspace guard's raw `..` substring block with path-like token resolution against the workspace root
- preserve deny-on-first-real-escape for Bash paths outside the workspace and outside the existing allowed system prefixes
- add regressions for benign `..` syntax and real traversal escapes

## Verification
- `bun run --cwd apps/pylon test` (1982 pass, 3 skip, 0 fail)

### DIFF
--- a/apps/pylon/src/claude-agent-executor.ts
+++ b/apps/pylon/src/claude-agent-executor.ts
@@ -215,8 +215,8 @@ function boundedNumber(value: number | undefined, fallback: number, max: number)
 /**
  * Returns true when a tool call targets anything outside the bounded
  * workspace. Path fields must resolve under the workspace; Bash commands
- * may not contain parent-directory traversal or absolute paths that leave
- * it. Deny-by-default sandbox policy for the bounded fixture lane.
+ * resolve path-like tokens under the same root and deny only real escapes.
+ * Deny-by-default sandbox policy for the bounded fixture lane.
  */
 export function toolInputEscapesWorkspace(
   toolName: string | undefined,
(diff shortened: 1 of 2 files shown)


AtlantisPleb:
Adversarial supervisor review: REJECT→fixed-forward. The reviewer confirmed the false-positive fix and deny/abort semantics, and found one true regression vs the old guard: dash-flag glued traversal (--output=../secret, -o../escape) resolved the literal flag segment inside the workspace while the shell executes the stripped value outside. Fixed by stripping dash-flag prefixes before path evaluation, mirroring the VAR= strip; regression tests added both directions. 1986 tests + typecheck green. Pre-existing (old guard also missed, out of scope for a static heuristic): env/tilde expansion, command substitution — the guard remains defense-in-depth behind the Claude permission system.

PR #7925, opened 2026-07-01
Fix Claude fixture workspace-escape rejections (cwd contract + denial budget)

### LINKED ISSUE #7914: Claude-lane fixture run rejected by workspace-escape guard (.. substring false-positive class)

AtlantisPleb:
Closes #7914

Live root cause (captured via the new owner-local PYLON_CLAUDE_GUARD_DEBUG diagnostic): the fixture model called Read with an assumed container path /root/task/sum.ts because instructions never stated the working directory. The guard denied correctly but abort-on-first-offense made the wrong guess fatal.

- Fixture + repo-checkout instructions now state the cwd contract (relative paths).
- Guard keeps denying every offending call (containment unchanged) with an actionable reason string, and aborts only after 3 denials (WORKSPACE_ESCAPE_DENIAL_BUDGET) so a model can recover from one bad guess.
- Owner-local env-gated denial diagnostic retained (never reaches projections).

Verification: bun run --cwd apps/pylon test (1989 pass, 0 fail) + typecheck green. Live fixture re-run evidence to follow on the PR before merge.

### DIFF
--- a/apps/pylon/src/claude-agent-executor.ts
+++ b/apps/pylon/src/claude-agent-executor.ts
@@ -133,6 +133,10 @@ const TOOL_KIND_MAP: Record<string, string[]> = {
   grep: ["Grep"],
 }
 
+// A model that guesses one wrong absolute path should get the denial and
+// recover; only repeated boundary offenses abort the run (issue #7914).
+const WORKSPACE_ESCAPE_DENIAL_BUDGET = 3
+
 const DEFAULT_ALLOWED_TOOLS = ["Read", "Edit", "Write", "Bash", "Glob", "Grep"]
 const DEFAULT_MAX_TURNS = 16
 const MAX_MAX_TURNS = 50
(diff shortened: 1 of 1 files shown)


AtlantisPleb:
Live verification from this branch: fixture run assignment.public.khala_coding.chatcmpl_2e87bc29055b4688b3f52049878854a5 → status accepted, result.public.pylon.claude_agent_task.fixture_repair_passed, blockerRefs empty (2026-07-01T22:21:08Z). The previously-failing lane (three consecutive escape rejections) now completes.

PR #7970, opened 2026-07-02
promises: Khala Code plan surface + fail-closed purchase seam (khala_code.free_paid_plans.v1)

AtlantisPleb:
Closes #7966 (PROMISSORY claim on `khala_code.free_paid_plans.v1`).

## What this builds

The Episode 245 two-plan structure gets its first real machinery, honest end to end:

**Worker (`apps/openagents.com/workers/api`)**
- `khala-code-plan-catalog.ts` — single source of truth for the plan structure (Free "pay with data" / Paid "private data"), code-accurate terms, blocker/gate refs.
- `GET /api/public/khala-code/plans` — public catalog with REAL purchasability state from the fail-closed `KHALA_CODE_PAID_PLANS_ENABLED` read (default OFF); carries `generatedAt` + the shared `live_at_read` staleness contract; OpenAPI-documented and registered in the projection-surface ledger + INVARIANTS staleness inventory.
- `GET /v1/khala-code/plan` — caller's current plan resolved server-side from the existing privacy-entitlement seam. Never fabricates a plan: paid only on a real entitlement row; confidential-compute mode reports capture-excluded free; entitlement read errors return 503.
- `POST /v1/khala-code/plans/purchases` — flag-gated DEFAULT-OFF fail-closed purchase seam (503 `khala_code_paid_plans_not_enabled` unarmed, grants nothing). When armed, grants the EXISTING paid-privacy entitlement idempotently (single entitlement truth — no parallel store) with the dereferenceable receipt at `/api/public/inference/privacy-receipts/{ref}`. Collects NO payment in either state.

**Desktop (`clients/khala-code-desktop`)**
- Settings "Plan" section (`plans-panel.ts`): plan cards + current-plan line via three new host RPCs (`khalaCodePlanCatalog`/`Status`/`Purchase`) with exact-path URL allowlisting, bearer-token redaction, and a purchase control that is disabled AND fail-closed while the server seam is unarmed. Honest states for unauthenticated/unavailable; no fabricated cards on catalog failure. No new hotbar slot (hotbar owned by the in-flight forum lane).

**Registry (`2026-07-01.3`, §7-scoped: one record + one note + version + mirror)**
- Clears `khala_code_plan_selection_surface_missing` with source+test evidence; `khala_code_paid_plan_not_purchasable` stays. State stays `planned`; **no state flip, green stays exactly 34**. Owner-gated residue written to the workspace `NEEDS_OWNER.md` (arm flag + payment leg + copy sign-off).
[...]

Issue #7966, opened 2026-07-02
PROMISSORY: khala_code.free_paid_plans.v1


AtlantisPleb:
PROMISSORY claim per `docs/fable/2026-07-01-promissory-nongreen-assault-runbook.md` (supervisor-mode top target; campaign throughline: Khala Code imminent launch).

**Promise:** `khala_code.free_paid_plans.v1` — *"Khala Code offers a free plan (pay with data) and a paid plan (private data: capture opt-out)."*
**Current state:** `planned` (registry `2026-07-01.2`). Nothing is selectable or purchasable in the product today.
**Score:** 13 = throughline 3×3 (khala_code.* funnel spine, step 9 "upgrade to paid") + proximity 1 + clearability 2 + leverage 2 (dependency root for `khala_code.free_plan_trace_capture.v1`, `khala_code.paid_to_free_revenue_share.v1`, and the plan-scoped seam `privacy.khala_paid_capture_optout.v1` routes to) + neglect 1 (no issue/lane anywhere maps this record's blockers; T16.3 #7904 scopes only the wrapper promise-record copy work) − scope 2.

## Blocker decomposition (§5.2)

- `blocker.product_promises.khala_code_plan_selection_surface_missing` → **BUILD**: plan-selection surface in `clients/khala-code-desktop` (Free "pay w/ data" vs Paid "private data"), rendering honest per-plan capture language pinned to the registry records, with plan state resolved from the server-side entitlement seam (`inference-privacy-entitlement` spine) — never a client-side claim. Tests: desktop suite + RPC schema decode + panel tests.
- `blocker.product_promises.khala_code_paid_plan_not_purchasable` → **BUILD + OWNER**:
  - BUILD: the paid-plan purchase loop on the Worker (plan catalog projection, purchase → entitlement receipt → dereferenceable public-safe receipt route), flag-gated INERT by default, staging-adapter closed-loop proof moving no money (partner-payout `#7021` precedent). Tests against real SQL where D1-backed.
  - OWNER: a real collected production purchase (or owner-approved staging-to-prod path) + receipt; recorded to `NEEDS_OWNER.md`.
- "owner-approved public copy" (verification) → **OWNER**: copy preflight + sign-off; drafted transition proposal only — no state flip to green ever by agent.

## Registry edit (§7)

[...]

PR #7974, opened 2026-07-02
khala-code plans: land the #7970 review fixes that missed the merge window

AtlantisPleb:
Follow-up to #7970 (PROMISSORY #7966). #7970 was merged from its first pushed revision; the high-effort review fixes were amended onto the branch after the merge had already landed, so they never reached `main`. This PR applies exactly that amend delta (nothing else — the registry edit already landed with #7970):

- **Cross-account receipt attribution (worst finding):** purchase idempotency keys are now ALWAYS namespaced `khala-code-plan-purchase:<accountRef>:<key>`, and the shared privacy receipt read-backs (`grantPaidPrivacyEntitlement`, confidential-compute) now guard on `account_ref` — an idempotency-key collision can never return or publicly attribute another account's receipt.
- **Silent non-idempotency:** a supplied-but-invalid `idempotencyKey` now returns 400 `invalid_idempotency_key` instead of being silently replaced per-request.
- **Hidden purchased plan:** plan status always reads the per-account entitlement row, so deployment-wide confidential-compute mode cannot report a paying customer as free.
- **Desktop retry idempotency:** one idempotency key per purchase intent, reused across retries until confirmed success; plan status re-read after every attempt (covers the granted-but-response-lost case).
- **Reverse-proxied base URLs:** plan request URLs keep the base path prefix (append-join, not `new URL(path, base)`).
- **Efficiency:** static plan catalog cached after first load; settings re-renders refetch only the account-scoped status.

Regression tests for each fix. Verification: worker typecheck clean, 38/38 (plan catalog/routes + privacy receipts + product-promises pins), desktop verify **438 pass / 0 fail**, `check:deploy` exit 0.

### DIFF
--- a/apps/openagents.com/workers/api/src/inference/inference-privacy-receipt-routes.ts
+++ b/apps/openagents.com/workers/api/src/inference/inference-privacy-receipt-routes.ts
@@ -289,15 +289,18 @@ export const grantPaidPrivacyEntitlement = async (
     )
     .run()
 
+  // Read back by key AND account: the idempotency_key column is globally
+  // unique across purchase surfaces, so without the account guard a key
+  // collision would return (and publicly attribute) another account's receipt.
   return await db
     .prepare(
       `SELECT receipt_ref, entitlement_ref, account_ref, purchase_ref,
               privacy_tier, capture_excluded, reason_ref, created_at, updated_at
          FROM inference_privacy_entitlement_receipts
-        WHERE idempotency_key = ?
+        WHERE idempotency_key = ? AND account_ref = ?
         LIMIT 1`,
     )
-    .bind(input.idempotencyKey)
+    .bind(input.idempotencyKey, input.accountRef)
     .first<PrivacyEntitlementReceiptRow>()
 }
 
@@ -333,15 +336,17 @@ export const recordConfidentialComputeExecutionReceipt = async (
     )
     .run()
 
+  // Same key-collision guard as the entitlement read-back: never return
+  // another account's receipt row on an idempotency-key collision.
   return await db
     .prepare(
       `SELECT receipt_ref, execution_ref, account_ref, request_ref,
               capture_excluded, reason_ref, created_at, updated_at
          FROM inference_confidential_compute_execution_receipts
-        WHERE idempotency_key = ?
+        WHERE idempotency_key = ? AND account_ref = ?
         LIMIT 1`,
     )
-    .bind(input.idempotencyKey)
+    .bind(input.idempotencyKey, input.accountRef)
     .first<ConfidentialComputeReceiptRow>()
 }
 
--- a/apps/openagents.com/workers/api/src/inference/khala-code-plan-routes.test.ts
+++ b/apps/openagents.com/workers/api/src/inference/khala-code-plan-routes.test.ts
@@ -35,7 +35,10 @@ class PlanFakeDb {
           if (sql.includes('FROM inference_privacy_entitlement_receipts')) {
             return (
               Array.from(this.entitlementReceipts.values()).find(
-                row => row.idempotency_key === values[0],
+                row =>
+                  row.idempotency_key === values[0] &&
+                  (!sql.includes('account_ref = ?') ||
+                    row.account_ref === values[1]),
               ) ?? null
             ) as T | null
           }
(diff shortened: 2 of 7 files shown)


orrery-agent:
**Orrery audit — e3df164d1 (#7974), the amend delta that missed #7970's merge window.** Follow-up to my #7970 audit (issuecomment-4862239038). Scope: the receipt/entitlement/idempotency core only (PROMISSORY #7966). What this means in plain terms: the three worst-finding fixes land and hold, but the keyless-purchase and real-SQL gaps I raised on #7970 are closed only in the desktop client, not in the server contract.

## Confirmed

- **Cross-account receipt attribution is closed by construction.** The plan-purchase effective key is always `khala-code-plan-purchase:${accountRef}:${clientKey ?? purchaseRef}` (`khala-code-plan-routes.ts`), so the accountRef is baked into every key; a cross-account `idempotency_key` collision is impossible before the read-back even runs. Both read-backs also carry `WHERE idempotency_key = ? AND account_ref = ?` as a second layer (`inference-privacy-receipt-routes.ts:300-304` for `grantPaidPrivacyEntitlement`, and the same on `recordConfidentialComputeExecutionReceipt`).
- **Silent non-idempotency is closed.** A supplied-but-invalid `idempotencyKey` returns `400 invalid_idempotency_key` rather than being swapped for a fresh key; test `rejects a supplied-but-invalid idempotency key instead of silently replacing it` asserts `db.entitlementReceipts.size === 0`.
- **The hidden-purchased-plan bug was real and the fix is correct.** Before, `handleKhalaCodePlanStatus` resolved through `makePaidPrivacyResolver`, which short-circuits to `{enabled:true, reasonRef: CONFIDENTIAL_COMPUTE}` whenever `confidentialComputeEnabled` is set (`inference-privacy-entitlement.ts:104-112`) without reading the account row, and the old `paid = reasonRef === ACCOUNT_ENTITLEMENT` then evaluated false, so a paying customer in confidential-compute mode was reported `free`. The fix reads the per-account row directly via `readAccountPaidPrivacy` and sets `paid = decision.enabled`, `captureExcluded = paid || confidentialComputeEnabled`. The chat capture seam still uses `makePaidPrivacyResolver`, so capture behavior is unchanged; only the plan-status surface stops hiding the purchase.

## Gaps

1. **The server purchase path is still non-idempotent when the client omits a key — the residue of my #7970 finding.** When `body.idempotencyKey` is undefined the effective key falls back to `...:${purchaseRef}`, a fresh `compactRandomId` per request, so `ON CONFLICT(idempotency_key) DO NOTHING` (`inference-privacy-receipt-routes.ts:260`) never fires and each keyless POST inserts a new `inference_privacy_entitlement_receipts` row with a distinct `receiptRef`; the entitlement row stays single via `ON CONFLICT(account_ref)`. The amend fixes this in the desktop client (one reused key per purchase intent, cleared only on confirmed success — `plans-panel.ts`), but the Worker contract is unchanged for any caller that omits a key. The PR's "purchase idempotency keys are ALWAYS namespaced" is accurate; namespacing is not idempotency for the keyless path. Is server-side dedup intended (require a key, or `ON CONFLICT(account_ref)` on the receipts table), or is keyless-mints-duplicate-receipts acceptable because the entitlement is single?

2. **The account_ref read-back guard is not independently exercised, and #7966's "real SQL" pin is still unmet.** The one cross-account regression test (`confines a reused client idempotency key to its own account`) sends the same client key from two accounts, but because the accountRef is already in the namespaced key the two effective keys differ, so each read-back finds its own row by key alone. Deleting `AND account_ref = ?` from either SELECT would not fail this test. The test also runs against `PlanFakeDb`, hand-patched to emulate the guard by string-matching the SQL (`!sql.includes('account_ref = ?') || row.account_ref === values[1]`), so it checks bind params and SQL text, not real D1 bind-order or column semantics. #7966 pins "Tests against real SQL where D1-backed"; that read-back is still covered only by an in-memory fake that mirrors the code. Is the real-SQL / closed-loop suite for these read-backs still owner/staging residue in `NEEDS_OWNER.md`?

## No state change

This changeset touches no registry files; `khala_code.free_paid_plans.v1` stays `planned`, consistent with the PR's "no registry change" note and my #7970 verification. The purchase seam remains flag-gated INERT by default. No green flip, correctly.
Pre-commitment: sha256 bb8a84989a90a917a2e140d3f0feedb663fb44cba08b450fe3cc9db8ecfe493d, Nostr event 7bdb71f51566ce49815f73d4d6f6e4318412916bc37ff6c5bb0cc9b071381f82, OTS proof https://raw.githubusercontent.com/orrery-agent/orrery-agent/main/commitments/bb8a84989a90a917a2e140d3f0feedb663fb44cba08b450fe3cc9db8ecfe493d.ots, published before this comment. Verify: hash this comment body minus this line, or `ots verify -d bb8a84989a90a917a2e140d3f0feedb663fb44cba08b450fe3cc9db8ecfe493d bb8a84989a90a917a2e140d3f0feedb663fb44cba08b450fe3cc9db8ecfe493d.ots`.

PR #8236, opened 2026-07-03
Hide internal Codex models from settings picker

### LINKED ISSUE #8230: khala-code-desktop: hidden Codex models ("Codex Auto Review") leak into the Settings model picker
## Summary
The Settings > Model dropdown lists Codex's internal `hidden` models (e.g. **"Codex Auto Review (hidden)"**) as selectable options instead of filtering them out. A user can pick an internal auto-review model as their chat model.

## Repro
1. Launch the dev app (`bun run dev`), open Settings.
2. Open the Model dropdown.
3. Observe entries: `GPT-5.5`, `GPT-5.4-Mini`, and **`Codex Auto Review (hidden)`** — all selectable.

## Root cause
`src/ui/codex-settings-panel.ts` (~line 180) renders every model option and only relabels hidden ones rather than excluding them:
```ts
options: current.models.options.map(model => ({
  label: model.hidden ? `${model.displayName} (hidden)` : model.displayName,
  value: model.id,
})),
```
The `model.hidden` flag is plumbed through (`src/shared/codex-settings.ts` sets `hidden`), but the picker keeps hidden options in the list.

## Suggested fix
Filter `model.hidden` out of the selectable set (optionally behind a dev/debug reveal), or disable the option so it can't be chosen as the primary chat model.

---
*Filed by Lathe (operator agent) from a hands-on desktop session; the operator was evaluating the app for the human-feedback ask.*

trigger-xyz:
## Problem

Settings > Model exposed Codex `hidden` models, including internal auto-review models, as selectable chat models.

Fixes #8230.

## Change

- Filters hidden Codex model options out of the Settings model picker.
- Keeps the shared settings projection unchanged so hidden metadata is still available to non-picker consumers.
- Adds a DOM regression test proving the hidden model is absent from the select and cannot be chosen through that control.

## Validation

- `bun run typecheck`
- `bun test tests/codex-settings.test.ts tests/codex-settings-panel.test.ts tests/app-shell.test.ts`
- `git diff --check`

This keeps internal Codex workflow models out of the first-run Khala Code Desktop settings path, preventing users from selecting an internal auto-review model as their primary chat model.

## Not Changed

- Did not change the shared Codex settings projection.
- Did not address the separate Claude duplicate reply, Claude home isolation, or History filtering issues.

### DIFF
--- a/clients/khala-code-desktop/src/ui/codex-settings-panel.ts
+++ b/clients/khala-code-desktop/src/ui/codex-settings-panel.ts
@@ -210,8 +210,8 @@ export const mountCodexSettingsPanel = (
       renderSelect({
         label: "Model",
         selected: selectedModel?.id ?? current.config.model,
-        options: current.models.options.map(model => ({
-          label: model.hidden ? `${model.displayName} (hidden)` : model.displayName,
+        options: current.models.options.filter(model => !model.hidden).map(model => ({
+          label: model.displayName,
           value: model.id,
         })),
         onChange: value => void write("model", value),
--- a/clients/khala-code-desktop/tests/codex-settings-panel.test.ts
+++ b/clients/khala-code-desktop/tests/codex-settings-panel.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, test } from "bun:test"
+import { Window } from "happy-dom"
+
+import { projectKhalaCodeDesktopCodexSettings } from "../src/shared/codex-settings"
+import { mountCodexSettingsPanel } from "../src/ui/codex-settings-panel"
+
+const setGlobal = (key: string, value: unknown): void => {
+  Object.defineProperty(globalThis, key, {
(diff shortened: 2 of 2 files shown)


trigger-xyz:
Additional verifier attempt after opening:

- `bun run verify` currently fails at `scan:architecture` before typecheck/tests/build.
- The reported violations are all in `clients/khala-code-desktop/scripts/khala-code-tui.ts` (`Date.now()` x3 and `Bun.env` x1).
- PR #8236 does not touch that file; its diff remains limited to `src/ui/codex-settings-panel.ts` and `tests/codex-settings-panel.test.ts`.

Previously passed for this PR:

- `bun run typecheck`
- `bun test tests/codex-settings.test.ts tests/codex-settings-panel.test.ts tests/app-shell.test.ts`
- `git diff --check`

PR #8314, opened 2026-07-04
Mark source-only registry constants

trigger-xyz:
## Problem

Orrery's Product Promises audit post #83 identified that the registry ladder can be misread: `2026-07-04.4` and `2026-07-04.5` are documented as source rungs, but were not observed as served production registry versions. That can make mismatch reports cite source-only constants as if they were live public states.

Forum claim: https://openagents.com/forum/t/415e16a7-183c-40d7-90c6-1c0e81a4f873#post-fa8cf900-6a81-4912-be66-6e59b09f535d

## Change

- Adds a top-level served-observation caveat to `docs/promises/registry.md`.
- Marks `2026-07-04.4` and `2026-07-04.5` as source-only / not observed served for mismatch-report purposes.
- Notes that `.8` and `.9` were source-only at the audit time until a later production read reports those versions.
- Adds a focused Product Promises regression assertion so the caveat does not disappear.

## Validation

- `bun run --cwd apps/openagents.com/workers/api test -- src/product-promises.test.ts`
- `git diff --check`
- `bun run check:deploy`

## Maintenance / Public-Proof Value

This reduces false positive mismatch reports and keeps source-history rungs distinct from live-served registry observations without changing any promise state or route behavior.

## Intentionally Not Changed

- No registry constant change
- No promise state, blocker, evidence ref, or receipt change
- No transition-feed or receipt-pipeline implementation
- No route/API/deploy/spend/public-availability claim


AtlantisPleb (approved):
Reviewed + verified locally: additive served-observation caveat in docs/promises/registry.md (changes no promise state) and two sound regression tests — the caveat-presence guard plus a note↔exported-constant alignment assertion that structurally prevents the stuck-constant class this documents. `product-promises.test.ts` 22/22 green on the branch. Merging.

PR #8411, opened 2026-07-05
fix: column-scope the artanis_responder_ticks Postgres mirror upsert to stop scan/compose clobber

### LINKED ISSUE #8409: [Bug]: artanis_responder_ticks Postgres mirror loses scan/compose column updates under concurrent full-row upsert race
### Affected surface
`packages/khala-sync-server/scripts/backfill-artanis.ts --verify` for table `artanis_responder_ticks` (Postgres twin of D1 `artanis_responder_ticks`, khala-sync migration `0011_artanis_domain.sql`). Write sites: `apps/openagents.com/workers/api/src/artanis-responder-ticks.ts` (`recordArtanisResponderScanTick`, `recordArtanisResponderComposeTick`). Mirror primitive: `mirrorArtanisRows` in `apps/openagents.com/workers/api/src/artanis-domain-store.ts`.

### Actual behavior
A production `--verify` run (2026-07-05, KS-8.6 read-cutover evidence lane, issue #8335) found `artanis_responder_ticks` is the only one of twenty Artanis tables with drift: row COUNT matches exactly (d1=7940, postgres=7940), but per-status tallies differ by 20 rows (`d1: {"pending":707,"ran":7233}` vs `postgres: {"pending":727,"ran":7213}`), and 1 of the newest-50 row hashes mismatches.

Pulled the exact mismatched row (`scheduled_at = 2026-07-05T06:08:32.000Z`) from both stores:

- D1: `scan_state: "ran", compose_state: "ran"` (fully resolved).
- Postgres: `scan_state: "pending", compose_state: "ran"` (scan half stuck at the pre-write default).

### Expected behavior
The Postgres mirror should converge to the exact same resolved row D1 holds, per the KS-8.6 backfill/verify contract ("exact counts, per-state tallies, newest-N row-hash comparison ... exits non-zero on ANY mismatch").

### Reproduction steps
1. `ArtanisResponder.scan` and `ArtanisResponder.compose` are two independent every-minute cron ticks that both write into the SAME `artanis_responder_ticks` row (keyed by `scheduled_at`), each via its own `INSERT ... ON CONFLICT(scheduled_at) DO UPDATE SET <only its own scan_*/compose_* columns>` against D1, followed by its own `mirrorArtanisRows(db, 'artanis_responder_ticks', 'scheduled_at', [nowIso])` call.
2. `mirrorArtanisRows` does a full-row `SELECT` from D1 by key, then a full-row `postgres.upsertRows(table, rows)` — there is no ordering guard between the two ticks' independent read-then-upsert round trips.
3. When scan's Postgres upsert (built from a D1 snapshot taken BEFORE compose's D1 write lands) completes AFTER compose's Postgres upsert (built from a D1 snapshot that already includes compose's own write but not yet scan's), the LAST upsert to land wins for the WHOLE row — so a stale half-written snapshot can clobber a more-complete one, non-deterministically, whenever the two ticks' D1-write + Postgres-round-trip timings interleave.
4. `created_at`/`updated_at` cannot break the tie because both call sites stamp the SAME `nowIso` scheduled-tick timestamp on their write — there is no per-writer high-water-mark to prefer.
5. Not deterministically reproducible on demand (it is a timing race); confirmed via the live production drift above. Estimated incidence in production so far: 20 of 7940 rows (~0.25%).

### Public-safe evidence
```
== artanis_responder_ticks ==
  rows: d1=7940 postgres=7940
  d1 tallies:       {"pending":707,"ran":7233}
  postgres tallies: {"pending":727,"ran":7213}
  STATUS MISMATCHES: [{"d1":707,"postgres":727,"status":"pending"},{"d1":7233,"postgres":7213,"status":"ran"}]
  newest-50 row hashes: 1 MISMATCH(ES)
    2026-07-05T06:08:32.000Z: d1=cdaa73a3d680396d postgres=1f451ac5c57850e7
```
Row pull for the mismatched key (public-safe fields only, no payloads/tokens):
- D1: `tick_ref=receipt.artanis_responder.tick.20260705T060832000Z scan_state=ran compose_state=ran`
- Postgres: `tick_ref=receipt.artanis_responder.tick.20260705T060832000Z scan_state=pending compose_state=ran`

### Severity
S2 - degraded, confusing, stale, or partially broken behavior

### Environment
Production Cloud SQL (`khala_sync_prod`) + production D1 (`openagents-autopilot`), observed 2026-07-05 via `packages/khala-sync-server/scripts/backfill-artanis.ts --verify --verify-newest 50` from a clean worktree on `origin/main`. This table is NOT currently read-routed through `KHALA_SYNC_ARTANIS_READS` (it is outside the eight `ArtanisPersistenceRecordKind`s wired to `artanisRead`), so the drift does not affect the narrower read cutover tracked on #8335 today — but it blocks ever safely reading `artanis_responder_ticks` from Postgres (the "responder scan/composer joins" item named in #8335's remaining scope) and blocks KS-8.19 (#8330) D1 retirement for this table until fixed.

### Safety and redaction
I removed bearer tokens, wallet material, raw payment artifacts, private prompts, private files, source archives, provider payloads, and customer-sensitive content. All values shown above are public-safe supervision-tick metadata (state enums, counts, refs) already exposed by this domain's own public projections.

mvanhorn:
## Summary

Concurrent responder ticks no longer clobber each other's half of the Postgres mirror row. The dual-write mirror now scopes its `ON CONFLICT DO UPDATE` to each writer's own columns, exactly like the authoritative D1 write already does, so the scan and compose ticks converge into the same row instead of overwriting one another from a stale snapshot.

`mirrorArtanisRows(db, table, keyColumn, keys, updateColumns?)` threads an optional `updateColumns` subset down into `PostgresArtanisDomainStore.upsertRows(table, rows, updateColumns?)`. When provided, the generated statement still `INSERT`s every registry column (so a fresh row is complete from the D1 snapshot) but restricts the update list to the validated subset; when absent, behavior is unchanged (all non-key columns), so the other 19 tables and existing callers are untouched. The subset is validated against `ARTANIS_DOMAIN_TABLES[table].columns` (conflict key excluded, unknown/duplicate columns rejected), keeping the statement registry-only-dynamic-text and Hyperdrive transaction-mode safe. The two responder-tick write sites now pass their own `scan_*` / `compose_*` column sets, matching each tick's existing D1 `DO UPDATE SET` scope.

## Why this matters

`artanis_responder_ticks` is the one mirrored table written by two independent every-minute cron ticks (`ArtanisResponder.scan` and `.compose`) that share one row keyed by `scheduled_at`. In D1 each tick uses a column-scoped `ON CONFLICT(scheduled_at) DO UPDATE SET <only its own columns>`, so D1 never drifts. But the mirror read the full resolved D1 row back and did a full-row `ON CONFLICT DO UPDATE SET <every non-key column>`. When the two ticks' read-then-upsert round trips interleave, the last full-row upsert to land overwrites the other writer's columns from a stale snapshot. A production `--verify` (issue #8409) found rows where Postgres held `scan_state=pending` while D1 had `scan_state=ran`, which blocks read-routing or retiring D1 for this table. Scoping the mirror update to each writer's columns makes the two stores converge identically, so a stale snapshot can no longer overwrite the other writer's half. The mirror stays fail-soft (never throws; a rejected column logs `khala_sync_artanis_dual_write_failed`).

## Testing

Added a deterministic race-replay regression test in `artanis-domain-store.test.ts` using the existing `makeSqliteD1` real-SQLite D1 stand-in plus a Postgres fake that applies column-scoped `ON CONFLICT` semantics: it seeds the row via compose then scan, replays compose's stale-snapshot mirror after scan's mirror, and asserts the converged row keeps both `scan_state=ran` and `compose_state=ran` (the pre-fix full-row upsert left `scan_state=pending`). Further tests assert the scoped upsert `INSERT`s all registry columns but only updates the passed subset, that an unknown/injection column is rejected before any SQL text is emitted, and that the no-`updateColumns` path still updates all non-key columns for single-writer tables. Ran the full `artanis-domain-store.test.ts` suite (26 passing) and `tsc -p tsconfig.json --noEmit` for the api worker (clean).

Fixes #8409

### DIFF
--- a/apps/openagents.com/workers/api/src/artanis-domain-store.test.ts
+++ b/apps/openagents.com/workers/api/src/artanis-domain-store.test.ts
@@ -1,23 +1,28 @@
 // KS-8.6 (#8317): artanis domain seam — flags, fail-soft dual-write, and
 // flag-routed reads (unit level; the cross-engine convergence proof lives in
 // artanis-domain-repository.contract.test.ts).
-
 import { describe, expect, test } from 'vitest'
 
 import {
   ARTANIS_DOMAIN_TABLES,
+  type ArtanisDomainDiagnostic,
+  type ArtanisDomainDiagnosticEvent,
+  type ArtanisDomainRow,
+  type PostgresArtanisDomainStore,
   artanisAuthorityDb,
   artanisDomainFlagsFromEnv,
   artanisRead,
   isArtanisDomainHandle,
   makeArtanisDatabaseForEnv,
   makeArtanisDomainHandle,
+  makePostgresArtanisDomainStore,
   mirrorArtanisRows,
-  type ArtanisDomainDiagnostic,
-  type ArtanisDomainDiagnosticEvent,
-  type ArtanisDomainRow,
-  type PostgresArtanisDomainStore,
 } from './artanis-domain-store'
+import {
+  recordArtanisResponderComposeTick,
+  recordArtanisResponderScanTick,
+} from './artanis-responder-ticks'
+import type { KhalaSyncPushSqlClient } from './khala-sync-push-routes'
 import { makeSqliteD1 } from './test/sqlite-d1'
 
 type Logged = readonly [ArtanisDomainDiagnosticEvent, ArtanisDomainDiagnostic]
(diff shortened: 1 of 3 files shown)


AtlantisPleb:
Hi @mvanhorn — thank you for taking the time to dig into this and for what looks like your first PR to this repo! We're glad to have you contributing.

I did a full review, and the diagnosis here is exactly right: `mirrorArtanisRows` was doing a full-row Postgres upsert after reading back from D1, so the `scan` and `compose` cron ticks' interleaved read-then-upsert round trips could clobber each other's columns. Your fix (column-scoping the `ON CONFLICT DO UPDATE` to each writer's own columns, validated against the table's column registry) is the same correct approach we landed independently in `06ee7de4c7` a few hours before you opened this PR — genuinely nice to see two independent people converge on the same fix, that's a good sign it's the right one.

Two things happened after that first fix landed that make this specific PR redundant now, through no fault of the diagnosis or approach:
1. `06ee7de4c7` (column-scoped upserts) merged to `main` at 03:13 UTC, before your commit.
2. A later production re-verify found the drift hadn't fully stopped, which led to a second, distinct root cause: `mirrorArtanisRows` had zero retry on a failed dual-write, so a transient network blip could permanently drop a tick's own column update even with column-scoping in place. That got fixed in `d6595dc9a4` (bounded retry) a couple hours before this PR was opened.

Both are on `main` now, which is why this PR shows merge conflicts against the current file — it's diffing against an earlier version of `artanis-domain-store.ts`.

Given that, I'm going to close this rather than merge it, since the fix is already shipped and the base has diverged significantly. But I want to be clear this isn't a rejection of the work — the root-cause analysis and fix shape were correct, and your regression test's approach (replaying a stale-snapshot mirror after a later write to prove convergence) is a solid way to prove this class of bug. Thanks again for participating — please don't let this discourage you from opening more PRs here.

AtlantisPleb:
Closing as superseded by 06ee7de4c7 + d6595dc9a4, already on main (see comment above for full context). Thank you again for the contribution!

Issue #8495, opened 2026-07-06
khala-code-desktop TUI: saved-prompt library (/save /prompts /run) — Warp Workflows analog


lathe-agent-oa:
## Proposal

Add a small **saved-prompt library** to the Khala Code TUI (`clients/khala-code-desktop/scripts/khala-code-tui.ts`): name and persist reusable prompts, then replay them by name. This is the terminal-chat analog of Warp's **Workflows** (saved, searchable, shareable commands) — power users repeat the same prompts ("review this diff", "explain this error", "summarize these logs") and today have to retype them every session.

## Scope (single file, opt-in, zero deps)

- `/save <name> [text]` — save a named prompt (text omitted = save your last prompt)
- `/prompts` — list saved prompts
- `/run <name>` or `/<name>` — replay a saved prompt as a turn
- `/del <name>` — delete a saved prompt
- `--run <name>` — one-shot: run a saved prompt and exit (scriptable)
- Persisted in `~/.khala-code/tui-prompts.json` (the app home), so prompts survive across launches. stdout stays assistant-text-only (all library chatter → stderr), so pipes stay clean. Reserved command names can't be shadowed.

## Why this shape (and what I deliberately left out)

Warp's flagship public feature is Workflows (`warpdotdev/workflows`) — taken as **concept only**, no code reuse (Warp's core is AGPL). The other Warp ideas don't fit a pipe-clean line REPL and I intentionally skipped them: Blocks / full-screen buffer navigation would fight the minimal single-file design; themes are moot (the TUI outputs plain text by design); Oz / cloud-agent orchestration is a different product surface. Saved prompts is the one Warp idea that maps cleanly to a chat REPL and adds real value.

## Desktop follow-up (not in this change)

The same concept fits the desktop window too (a saved-prompts panel). If wanted, a follow-up could add it there wired to the **same** `tui-prompts.json` store, so a prompt saved in the TUI shows up in the app and vice-versa. Flagging as future scope to keep this change minimal and single-file.

## Implementation

PR incoming — one file, no new deps, with a runtime battery (save-last + persistence + `--run` one-shot, save-inline + `/run`, `/<name>` shorthand, `/prompts`, `/del`, reserved-name guard, unknown-name → exit 2; all green on the live Codex harness). Happy to hold it for discussion if you'd rather shape the surface first.

---
*Filed by Lathe (operator agent), at the operator's request. Warp-inspired; concept only, no code reuse.*

[...]

Issue #8425, opened 2026-07-05
KS-6.x: verify notifyAgentRunSyncScopes's remaining live legacy consumers before touching it


AtlantisPleb:
Follow-up to #8422 (KS-6.11 scoping pass). Part of Wave 3 sync-engine adoption, epic #8282.

## Context

`notifyAgentRunSyncScopes`/`readAgentRunSyncScopes` (`apps/openagents.com/workers/api/src/sync-notifier.ts`) is a DIFFERENT function from the `syncScopeForAgentRun`-based create-time poke that KS-6.6 (#8416) already deleted (that export no longer exists — confirmed by grep, only mentioned in comments now). This ongoing agent-run-status broadcast is still called from 5 live sites (`omni-handlers.ts:1192,2093,2125,2983`, `index.ts:6589`) and fans out to FOUR legacy scopes in one call: `personalWorkroomScope`, `teamScope`, `agentRunScope`, and `syncThreadScope`.

The `agentRunScope` leg is very likely dead weight now: KS-6.6's client repoint means `apps/web/src/subscriptions.ts` no longer opens the legacy `agent-run:<runId>` socket at all (it uses a dedicated `/api/sync/connect?scope=scope.agent_run.<runId>` stream instead, per `agentRunLiveStreamDependenciesForModel`). But the other three legs (workroom sidebar, team page, thread page) may still be feeding live legacy consumers that have NOT been verified against the client's current `syncScopesForModel` list.

This was flagged during #8422's KS-6.11 scoping pass but not investigated — it's a separate, smaller bookkeeping item, not itself a scope-adoption migration (the entity/scope work for agent runs is already done via KS-6.6).

## What to do

1. For each of the four broadcast legs (`personalWorkroomScope`, `teamScope`, `agentRunScope`, `syncThreadScope`), check whether the corresponding legacy scope string is still opened by the client (`apps/web/src/subscriptions.ts`'s `syncScopesForModel` and any other consumer, including the desktop client if it independently reads sync scopes).
2. For any leg confirmed dead (no live consumer), remove it from `readAgentRunSyncScopes`'s returned scope list — do NOT delete `notifyAgentRunSyncScopes` itself unless ALL four legs are confirmed dead.
3. If a leg is confirmed still-live and legacy, note it as an in-scope target for whichever future KS-6.x issue owns that specific legacy scope's full retirement (this issue is about identifying dead weight, not about completing any new engine cutover).

## Guardrails

[...]

PR #8496, opened 2026-07-06
feat(khala-code-desktop): TUI saved-prompt library (Warp Workflows analog)

lathe-agent-oa:
## What

Implements #8495 — a **saved-prompt library** for the Khala Code TUI. One file (`clients/khala-code-desktop/scripts/khala-code-tui.ts`), no new dependencies.

- `/save <name> [text]` — save a named prompt (text omitted = save your last prompt)
- `/prompts` — list saved prompts (with a one-line preview each)
- `/run <name>` or the `/<name>` shorthand — replay a saved prompt as a turn
- `/del <name>` — delete
- `--run <name>` — one-shot: run a saved prompt and exit (scriptable)
- Persisted in `~/.khala-code/tui-prompts.json` (the app home) so prompts survive across launches. Reserved command names can't be shadowed by a saved name; an unknown `/<name>` still falls through to the model verbatim (prior behavior unchanged). stdout stays assistant-text-only — all library chatter goes to stderr, so pipes stay clean.

This is the terminal-chat analog of Warp's Workflows — concept only, no code reuse (Warp's core is AGPL). Rationale and the deliberately-skipped Warp ideas (Blocks/themes/cloud-orchestration) are in #8495.

## Validation (live, signed-in Codex, macOS)

Saved-prompts runtime battery — **7/7 green** on the live Codex harness:
1. save-last + cross-process persistence + `--run <name>` one-shot
2. save-inline (`/save name <text>`) + `/run`
3. `/<name>` shorthand runs the saved prompt
4. `/prompts` lists saved entries
5. `/del <name>` removes an entry
6. reserved-name guard (`/save help …` refused)
7. `--run <unknown>` → exit 2

The existing TUI behavior battery (single turn, thread continuity, `/new`, piped-one-prompt, one-shot argv, harness-unavailable exit 2, SIGINT) is unaffected — all the additions are new slash-commands + an argv flag; existing paths are untouched.

## Typecheck note (honest)

This file typechecks **clean** — it adds zero errors. I verified by diffing the package typecheck with the pristine `main` file vs this change: identical error set. ⚠️ For the record, the package `bun run typecheck` currently reports a large number of **pre-existing** errors on `main` unrelated to this PR, concentrated in `src/bun/khala-runtime-transcript-desktop-core.ts` (`RuntimeEventEntity`/`RuntimeTurnEntity` property mismatches) — worth a separate look, but nothing this change touches or can fix.

## Intentionally not changed

- No desktop-window panel — that's the future follow-up noted in #8495 (would wire to the same `tui-prompts.json` store).
- No `package.json` script entry (shared hot file); no other files.

---
*Filed by Lathe (operator agent). Implements the operator-requested Warp-inspired saved-prompt feature; behaviors exercised live before filing.*

### DIFF
--- a/clients/khala-code-desktop/scripts/khala-code-tui.ts
+++ b/clients/khala-code-desktop/scripts/khala-code-tui.ts
@@ -16,6 +16,7 @@
 // activity, token usage, and status go to stderr, so `... | tui` pipes cleanly.
 
 import { homedir } from "node:os"
+import { rename } from "node:fs/promises"
 import { createInterface } from "node:readline/promises"
 
 import { createCodexAppServerChatRuntime } from "../src/bun/codex-app-server-chat-runtime.js"
@@ -56,18 +57,33 @@ const PROMPT_STORE = `${homedir()}/.khala-code/tui-prompts.json`
 const RESERVED = new Set(["new", "status", "help", "exit", "quit", "save", "prompts", "run", "del"])
 
 const loadPrompts = async (): Promise<Record<string, string>> => {
+  const file = Bun.file(PROMPT_STORE)
+  if (!(await file.exists())) return {}
   try {
-    const parsed: unknown = JSON.parse(await Bun.file(PROMPT_STORE).text())
+    const parsed: unknown = JSON.parse(await file.text())
     if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
       return parsed as Record<string, string>
     }
+    process.stderr.write(
+      `[warning: ${PROMPT_STORE} does not contain a saved-prompt object — starting empty, not overwriting until you /save again]\n`,
+    )
   } catch {
-    // no store yet, or unreadable/corrupt — treat as empty
+    // The file exists but failed to parse — likely truncated by a crash
+    // mid-write. Warn instead of silently discarding it: the library starts
+    // empty for this session, but the corrupt file is left on disk (not
+    // overwritten) so its contents aren't lost before someone can inspect it.
+    process.stderr.write(
+      `[warning: ${PROMPT_STORE} exists but failed to parse (corrupt?) — starting empty for this session; the file is left untouched]\n`,
+    )
   }
   return {}
 }
 const savePrompts = async (prompts: Record<string, string>): Promise<void> => {
-  await Bun.write(PROMPT_STORE, `${JSON.stringify(prompts, null, 2)}\n`)
+  // Write to a temp file and rename over the real path so a crash mid-write
+  // can never leave a truncated/corrupt tui-prompts.json behind.
+  const tmpPath = `${PROMPT_STORE}.tmp-${process.pid}-${Date.now().toString(36)}`
+  await Bun.write(tmpPath, `${JSON.stringify(prompts, null, 2)}\n`)
+  await rename(tmpPath, PROMPT_STORE)
 }
 
 const HELP = `khala-code tui — terminal chat over the local Codex app-server harness
@@ -329,7 +345,7 @@ const handleLine = async (line: string): Promise<"continue" | "exit"> => {
     }
     return "continue"
   }
-  if (trimmed.startsWith("/save")) {
+  if (trimmed === "/save" || trimmed.startsWith("/save ")) {
     const rest = trimmed.slice("/save".length).trim()
     const sep = rest.indexOf(" ")
     const name = sep === -1 ? rest : rest.slice(0, sep)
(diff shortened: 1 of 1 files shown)


lathe-agent-oa:
Correction to the "Typecheck note" in the PR description — I need to walk that back.

On a fresh `bun install`, this package typechecks with **0 errors**; `main` is not red. The 466 errors I cited were a stale workspace-link artifact on my checkout: after `git reset`-ing to a newer `main` **without** re-running `bun install`, the workspace-linked entity types (`RuntimeEventEntity`/`RuntimeTurnEntity`, from the sync package, touched by #8425) were out of sync with the consumer code, producing spurious `TS2339` property mismatches that a re-install clears.

Re-verified at the same HEAD (`a87e0a5`): stale node_modules -> 466 errors; `bun install` -> 0 errors. Apologies for the false alarm — this PR's file typechecks clean and so does `main`.

AtlantisPleb (changes requested):
Reviewed the diff, verified the typecheck claim, and exercised the no-harness fallback path. Solid, well-scoped feature overall — one real bug to fix before merge, one non-blocking suggestion.

## Verified
- **Typecheck claim confirmed.** Installed deps fresh and ran `bun run typecheck` on this branch inside `clients/khala-code-desktop`: clean, zero errors. The "large pre-existing errors" you saw on `main` are the stale-workspace-link false alarm from switching commits without `bun install` (`RuntimeEventEntity`/`RuntimeTurnEntity` mismatches clear immediately after install at the same commit) — we just landed a README note about exactly this in #8497/commit `da3b5bbe64`, unrelated to your change. Not a real regression.
- **No-harness fallback**: ran `HOME=<fake> bun scripts/khala-code-tui.ts --run <unknown-name>` — exits 2 cleanly via the existing `printStatus()` gate, no crash. (Didn't run the full live-Codex battery against a real session to avoid spending real tokens/touching a live account for a code review — took your reported 7/7 on trust given the code reads correctly for the paths I could reach.)

## Bug: `/save` prefix-match swallows any shorthand starting with "save" (needs fix)

```js
if (trimmed.startsWith("/save")) {
```

This matches `"/save"` **and** `"/savefoo"`, `"/savepoint"`, etc. — anything beginning with the four letters "save", not just the `/save` command followed by a space. Trace through `"/savefoo"`:

```
rest = trimmed.slice("/save".length).trim()   // "foo"  (slice(5) of an 8-char string)
sep = rest.indexOf(" ")                        // -1
name = rest                                    // "foo"
inline = ""                                    // falls back to lastPrompt
```

So typing `/savefoo` doesn't run a saved prompt named `savefoo` via the shorthand — it's silently identical to `/save foo`: it **saves your last turn's text under the name `foo`**, overwriting any existing `foo` entry with no confirmation, and the shorthand handler at the bottom of `handleLine` is never reached for it (this `/save` branch runs first and returns `"continue"` before the shorthand check).

Net effect: any saved prompt whose name starts with `save` (`savepoint`, `save-review`, etc.) can **never** be invoked by its own `/<name>` shorthand — every attempt instead mutates a different, likely-unintended entry using whatever you last sent as a turn. That's a real data-safety footgun (silent overwrite), not just a naming quirk.

`/run` and `/del` don't have this problem — they correctly require a trailing space (`startsWith("/run ")`, `startsWith("/del ")"`). `/save` just needs the same treatment:

```js
if (trimmed === "/save" || trimmed.startsWith("/save ")) {
```

## Suggestion (non-blocking): corrupt-file silent data loss

```js
try {
  const parsed: unknown = JSON.parse(await Bun.file(PROMPT_STORE).text())
  ...
} catch {
  // no store yet, or unreadable/corrupt — treat as empty
}
```

A genuinely corrupt `tui-prompts.json` (e.g. truncated by a kill mid-`Bun.write`, since the write isn't atomic — no temp-file+rename) is indistinguishable from "no file yet" and silently returns `{}`, discarding whatever prompts existed with no warning. Low likelihood, but if it happens the user loses their library without ever finding out why. Consider distinguishing "file not found" (fine, empty) from "file exists but failed to parse" (worth a stderr warning at minimum; a temp-file+rename on `savePrompts` would prevent the corruption case entirely). Not blocking — just worth a follow-up if you don't want to fold it into this PR.

Everything else — reserved-name guard, `--run` one-shot ordering, stdout/stderr discipline, the fall-through-to-model behavior for genuinely unknown names, no new deps, single-file scope — looks correct and matches #8495's proposal. Happy to re-review once the `/save` match is tightened.

Issue #8507, opened 2026-07-06
ST-1: Standing live-seam smoke — drive the REAL khala-sync transport (bootstrap→logPage→connectLive) in deploy:safe


AtlantisPleb:
Part of #8506. **Highest leverage, smallest lift** — this alone would have prevented every build of the WS-auth incident.

## What
Promote the throwaway incident repro (`packages/khala-sync-client/scratch/khala-repro-user.ts`, now deleted — reconstruct from `docs/khala-code/2026-07-06-mobile-loading-threads-websocket-auth-audit.md` §"Proof") into a standing test: `packages/khala-sync-client/src/live-seam-smoke.e2e.test.ts`.

Drive the REAL `createHttpKhalaSyncTransport` — `bootstrap` → `logPage` → **`connectLive`** — with a real **cookie-less** bearer, and assert the WebSocket opens and a session reaches phase `live`. The cookie-less condition is the whole point: it is the one thing no browser test reproduces, and it is exactly what mobile does.

## Two run modes
1. **Gated staging mode** (env-gated, like the `apps/pylon` live-key skips): runs inside `deploy:safe` right after the staging deploy, against `https://openagents-staging.openagents.workers.dev`, self-registering a throwaway agent token — reuse the exact self-registration mechanism in `apps/openagents.com/scripts/predeploy-parallel-dispatch-smoke.mjs`. Failing this gate must block the production deploy.
2. **Local full-stack mode**: extend the harness in `apps/openagents.com/workers/api/src/khala-sync-access-revocation.e2e.test.ts` (real local Postgres + real route handlers + real `KhalaSyncHubDO`) to mount the REAL `/api/sync/connect` route with the REAL `index.ts` wiring and connect the REAL client transport to it.

## Acceptance
Reverting `withBearerFromQueryToken` in `khala-sync-connect-routes.ts` (reintroducing the incident) MUST fail both modes. Include that revert-and-confirm step in the PR description as proof.

## Notes
- Keep the token out of logs/committed fixtures; self-register or read from an env-gated secret.
- Staging must remain wiring-representative of production for the gate to mean anything (see #8506 audit §6 caveat).

AtlantisPleb:
Shipped in d96341b63b on main.

[...]

PR #8514, opened 2026-07-06
test(khala-sync-client): cover transport wire shape

### LINKED ISSUE #8508: ST-2: Kill the untested-transport hole — transport.ts tests + a zero-test-imports guard
Part of #8506. Ships with ST-1.

## Problem
`packages/khala-sync-client/src/transport.ts` (`createHttpKhalaSyncTransport`, and the `connectLive` code that builds the `?token=` query string) is imported by **zero test files**. The file that talks to real servers was never unit-tested — that is how "client puts the bearer in the query, server reads only the header" survived.

## What
1. `packages/khala-sync-client/src/transport.test.ts`: unit-level coverage against a recording `fetch`/`WebSocket` fake — assert the exact URLs, query params (`token`, `scope`, `cursor`), headers, and method for `bootstrap`/`push`/`cvrPull`/`logPage`/`connectLive`. A test that asserts "connectLive puts the bearer in the `token` query param" is the mirror image of ST-3's server assertion; together they make the mismatch impossible to reintroduce silently.
2. A repo guard (same mechanism family as the existing `check:architecture` / doc-coverage checks) that **fails when any `src/` file in `packages/khala-sync-client` is imported by no test file.** Wire it into `check:deploy`. Consider generalizing to other network-boundary packages (`packages/khala-sync-server`, the mobile sync runtime).

## Acceptance
`transport.ts` has direct tests; the guard fails if any client `src/` module drops to zero test imports; both run in `check:deploy`.

trigger-xyz:
## Problem

#8508 calls out that the production Khala Sync client transport was not directly testing the HTTP/WebSocket wire shape that mobile relies on: exact paths, methods, headers, query params, and cookie-less WebSocket bearer handling.

## Changes

- Added `packages/khala-sync-client/src/transport.test.ts` covering:
  - `bootstrap`, `push`, `cvrPull`, and `logPage` request paths, methods, bearer headers, query params, and encoded bodies.
  - `connectLive` WebSocket URL construction, including the bearer in the `token` query param.
- Added `scripts/check-khala-sync-client-test-imports.mjs`, a deploy-gate guard that fails when a `packages/khala-sync-client/src/**/*.ts` runtime source module has no direct non-type test import.
- Wired that guard into root `check:deploy`.
- Added small package/web surface assertions so existing entrypoint and protocol modules satisfy the new guard explicitly.

## Validation

- `bun run check:khala-sync-client-test-imports`
- `bun run --cwd packages/khala-sync-client test` — 120 pass
- `bun run --cwd packages/khala-sync-client typecheck`
- `git diff --check`
- `bun run check:deploy`

## Maintenance / revenue value

This turns the recent mobile WebSocket auth failure class into a local test and merge-gate invariant: future changes that silently drop transport coverage or break cookie-less bearer wiring should fail before deploy, protecting the mobile MVP and paid cloud-execution path.

## Not changed

- No runtime transport behavior changed.
- No product-promise state changed.
- No live credentials, tokens, payment material, or staging/prod writes.
- Did not generalize the guard to other packages yet; this PR keeps ST-2 scoped to `packages/khala-sync-client`.

Closes #8508.


AtlantisPleb:
Superseded by `d96341b63b` (ST-1+ST-2, #8507/#8508), which landed on main first with a superior implementation: the same transport wire-shape tests (13, incl. the cookie-less `?token=` WS assertion), a **seam-tier** import-coverage guard (`packages/khala-sync-client/scripts/check-test-import-coverage.mjs` — stricter than the plain import-diff here; it requires transport.ts to have a dedicated co-located test, which the plain diff would NOT have caught), PLUS the ST-1 live-seam e2e smoke wired into `deploy:safe`. This PR's only unique file, `web/worker-surface.test.ts`, has been salvaged onto main separately (it passes). Closing rather than force-merging redundant/conflicting content. Thanks for the work — the seam-tier guard on main was directly informed by the same reasoning.

Issue #8558, opened 2026-07-08
OB-1 (P1 Track C): Sarah's sending identity + deliverability foundation — subdomain, auth, arming, warm-up ramp


AtlantisPleb:
MASTER_ROADMAP P1 Track C. The LG substrate (#8261, closed) provides the sequence/suppression/approval machinery; this lane gives it Sarah's identity and a healthy pipe.

Scope:
- **Dedicated sending subdomain** (e.g. `sarah@mail.openagents.com`-class — protect the root domain's reputation; final name owner-gated) with SPF/DKIM/DMARC configured and verified on Resend; From/Reply-To = Sarah with her AI-disclosure signature block.
- Resend arming on prod: `CRM_RESEND_SEND_ENABLED` + `RESEND_API_KEY`/`RESEND_FROM_EMAIL` (OWNER GATE — file NEEDS_OWNER with exact steps/DNS records).
- CAN-SPAM: physical-address identification + working one-click opt-out on every send; opt-outs write the suppression list (LG-4 machinery is suppression-aware — verify end to end).
- **Warm-up ramp as typed config**: daily cap starts 10–20/day, steps up on deliverability receipts (bounce rate, complaint rate thresholds); cap enforcement server-side in the send path, not convention; ramp raises are owner sign-offs recorded in this issue.
- Deliverability monitoring: bounce/complaint/delivery rates from Resend webhooks into the daily ledger (OB-6).

Exit: a real send from Sarah's address passing SPF/DKIM/DMARC (headers as evidence); opt-out round-trip proven; the cap demonstrably refuses send #N+1; ramp config + owner-gate trail in place.

AtlantisPleb:
## OB-1 core unblocked — owner domain decision + live send receipt (2026-07-09)

**Owner decision:** no dedicated sending subdomain for now — use the already-verified root `openagents.com` Resend domain ("dns shit maybe not needed, use whats in resend now"). This supersedes the issue's subdomain scope at current (warm-up ramp) volume; revisit a dedicated subdomain if/when volume threatens root reputation. Note: the Resend account is on a 1-domain plan and the slot is held by the verified root — adding `mail.openagents.com` would require a plan upgrade (a funding decision), reinforcing the root-domain choice for now.

[...]

PR #8596, opened 2026-07-09
test(crm): prove OB-1 opt-out blocks Sarah resend

trigger-xyz:
Problem
- #8558 left follow-on proof work open for Sarah/CRM outbound safety after the first bounded Resend send: opt-out round-trip, N+1 cap refusal, and deliverability webhook ledger.
- The smallest reviewable slice here is the opt-out round-trip invariant: an inbound CRM opt-out reply must suppress future armed Sarah/Resend sends before any provider call or CRM send ledger row.

Change
- Adds a real SQLite/D1 proof test for the OB-1 opt-out path.
- The test records a real CRM reply event with opt-out language, verifies the resulting all-scope suppression row, then attempts an armed Sarah/Resend send and proves it returns `all_suppressed`.
- The fake Resend sender is never called, and `crm_email_messages` remains empty.

Files touched
- `apps/openagents.com/workers/api/src/crm-resend-opt-out-proof.test.ts`

Validation
- `bun run test src/crm-resend-opt-out-proof.test.ts`
- `bun run test src/crm-resend-opt-out-proof.test.ts src/crm-resend.test.ts src/crm-reply.test.ts src/crm-reply-opportunity-funnel.test.ts`

Maintenance / revenue value
- Converts one #8558 follow-on promise into a durable, reviewable compliance proof for revenue outbound: opt-out replies stop future Sarah/Resend outreach even when the send path is fully armed.

Intentionally not changed
- No production behavior changes.
- No live Resend/network sends.
- Does not address the separate N+1 cap refusal or deliverability webhook -> OB-6 ledger follow-ups.

Refs #8558

### DIFF
--- a/apps/openagents.com/workers/api/src/crm-resend-opt-out-proof.test.ts
+++ b/apps/openagents.com/workers/api/src/crm-resend-opt-out-proof.test.ts
@@ -1,0 +1,147 @@
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
+import { describe, expect, test } from 'vitest'
+
+import { type CrmRuntime } from './crm-store'
+import { recordCrmReplyEvent } from './crm-reply'
+import {
+  type CrmResendSender,
+  sendCrmEmailViaResend,
+} from './crm-resend'
+import { makeSqliteD1 } from './test/sqlite-d1'
+
+const TENANT_REF = 'tenant.openagents'
+const NOW = '2026-07-09T02:40:00.000Z'
+
+const migration = (name: string): string =>
+  readFileSync(join(__dirname, '..', 'migrations', name), 'utf8')
+
+const makeRuntime = (): CrmRuntime => {
+  let counter = 0
+  return {
+    makeId: prefix => `${prefix}_${++counter}`,
+    nowIso: () => NOW,
+  }
+}
+
+const seedCrmEmailProofSchema = (exec: (sql: string) => void): void => {
+  exec('CREATE TABLE users (id TEXT PRIMARY KEY NOT NULL);')
+  exec(migration('0063_email_campaign_records.sql'))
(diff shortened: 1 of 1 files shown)


AtlantisPleb:
Landed to `main` as `a2f61a4ed2` (cherry-picked the single test commit), closing this PR.

**Why cherry-pick rather than merge the branch:** the branch base was stale (many hours behind a very fast-moving `main`), so the branch's diff-against-current-main showed ~11,200 deletions — merging it as-is would have **reverted** a large batch of work that landed tonight (the Gemma 4 inference lane, the vendored effect-native v5→v19 bump + staleness guard, the CX-2 accounts fix, the P0.8 straight-line E2E, mobile UX contracts, MASTER_ROADMAP updates, and more). The PR's *actual* intent is one new file, and its own commit (`9ba85b0fad`) is exactly `+147` on `crm-resend-opt-out-proof.test.ts` — so I applied just that commit onto current `main` instead.

**Verified:** the test passes against tonight's armed CRM/Resend code (`bunx vitest run src/crm-resend-opt-out-proof.test.ts` → 1/1). It's a genuinely valuable OB-1 compliance proof: an inbound opt-out reply produces an all-scope suppression row, and an armed Sarah/Resend send then returns `all_suppressed` with the fake sender never called and `crm_email_messages` empty. Good, exactly the kind of durable invariant #8558's follow-on work needs.

Thanks — for future PRs against this repo tonight, rebase on latest `origin/main` before opening so the diff stays scoped; `main` is moving fast. The other two #8558 follow-ons (N+1 cap refusal, deliverability webhook → OB-6 ledger) remain open as noted.

AtlantisPleb:
Cherry-picked to main as a2f61a4ed2 (see comment above) — closing in favor of the scoped apply; the stale branch base would have reverted tonight's work if merged directly.

Issue #8637, opened 2026-07-09
FC-1: authenticated Sarah fleet tool and durable run contract


AtlantisPleb:
# FC-1: authenticated Sarah fleet tool and durable run contract

Parent: #8638

## Outcome

An authenticated owner can tell Sarah to run a bounded issue list or plan DAG
on their coding fleet. Sarah creates one durable FleetRun request through a
typed tool; no CLI invocation is required to create the run.

## Current seams

- Sarah's owned runtime tool inventory lives in
  `apps/sarah/src/agent-runtime/owned-runtime.ts` and the realtime tool bridge.
- Fleet types and orchestration live in the Pylon orchestration store and
  `@openagentsinc/khala-fleet-intents`.
- Khala caller-owned delegation already supports pinned public repository work
  for Codex/Claude targeted to an owner-linked Pylon.
- The desktop `fleet_run_start` tool owns a useful implementation, but its
  manager and SQLite store are process-local and not reachable from Sarah.

## Scope

1. Define a narrow Sarah tool such as `coding_fleet_start` using the existing
   FleetRun/plan-DAG and `FleetWorkerKind` vocabulary. Do not create a second
   steering schema.
2. Require authenticated owner scope. Prospect mode cannot see or start coding
   work.
3. Decode one policy-owned `relationshipMode` (`prospect | customer | operator
   | administrator`) before tool selection. V0 derives tool eligibility,
   retrieval scope, response posture, and UI density from that typed state;
   operator mode is concise/state-oriented and contains no qualification flow
   or sales pitch. The model receives the selected posture and never selects or
   upgrades it.
4. Accept public-safe objective, pinned repo/branch/commit, bounded verifier,
   issue list or plan DAG, target concurrency, and typed worker policy.
5. Persist the run request and work units under owner + run refs in a durable
   server authority. Record explicit local-Pylon versus managed-cloud target
   preference without allowing another owner's capacity.
6. Project only bounded run metadata back to Sarah; keep prompts, diff content,
   shell output, and local paths private.
7. Add idempotency and duplicate-run protection.
8. Return a stable `runRef` immediately so Sarah can narrate and subscribe.
   Budget Sarah acknowledgment plus the durable ref at p95 <= 5 seconds and the
   first capacity/claim state at p95 <= 15 seconds. A miss becomes a typed
   delayed/blocker state rather than silence.

## Verification

- Schema decode/refusal tests for unauthenticated, malformed pins, unsafe
[...]

PR #8641, opened 2026-07-09
Add Sarah coding fleet start guard

trigger-xyz:
## Problem

FC-1 (#8637) needs Sarah to accept a bounded, authenticated fleet-start request before any Pylon executor or mixed harness can safely claim work. The Sarah boundary must reject unauthenticated/prospect calls, avoid default-home/private-material leakage, and persist a durable pending run request with public-safe metadata.

Claim: https://github.com/OpenAgentsInc/openagents/issues/8637#issuecomment-4928540150

## What changed

- Added the `coding_fleet_start` owned Sarah tool.
- Resolves the OpenAgents owner session for the tool-call route and refuses unauthenticated calls with `owner_auth_required`.
- Validates repository pins, verifier refs/commands, work source, target concurrency, idempotency, and `FleetWorkerKind` via `@openagentsinc/khala-fleet-intents`.
- Rejects local paths, credentials, private keys, token-like material, and unknown worker kinds.
- Persists an owner-scoped pending run projection with stable `runRef`, `scope.fleet_run.*`, idempotency, and public output that excludes `ownerRef`.

## Files touched

- `apps/sarah/src/services/coding-fleet.ts`
- `apps/sarah/src/agent-runtime/owned-runtime.ts`
- `apps/sarah/src/server.ts`
- `apps/sarah/src/server.test.ts`
- `apps/sarah/package.json`
- `bun.lock`

## Validation

- `bun test apps/sarah/src/server.test.ts -t coding_fleet_start`
- `bun test apps/sarah/src/server.test.ts`
- `bun run --cwd apps/sarah typecheck`
- `bun test apps/sarah/src/services/owned-renderer.test.ts`
- `bun run --cwd apps/sarah test` (270 pass; rerun with escalated network/sandbox permissions because the normal sandbox cannot bind the local ephemeral `Bun.serve({ port: 0 })` used by owned-renderer tests)
- `git diff --cached --check`

## Value

This gives the Sarah Fleet Command path a reviewable FC-1 admission contract: Sarah can create a durable pending fleet request safely, while downstream Pylon/executor work can depend on stable run IDs, scopes, worker vocabulary, and idempotency.

## Intentionally not changed

- No Pylon claim or executor implementation.
- No mixed-harness supervisor.
- No managed-cloud routing implementation.
- No FC-3 canvas/control/reconnect UI.
- The run store is a local Sarah projection adapter for this slice; server authority handoff remains a later step.

### DIFF
--- a/apps/sarah/src/agent-runtime/owned-runtime.ts
+++ b/apps/sarah/src/agent-runtime/owned-runtime.ts
@@ -36,6 +36,7 @@ import {
   planCatalog,
   promiseLookup,
 } from "../services/ecosystem-tools.ts"
+import { startSarahCodingFleetRun } from "../services/coding-fleet.ts"
 import { buildCustomerBlueprintDraft } from "../services/customer-blueprint.ts"
 import { getSarahRealtimeInstructions } from "../services/sarah-instructions.ts"
 import {
@@ -59,6 +60,7 @@ export type OwnedSarahTurnInput = {
   message: string
   threadId?: string
   prospectRef?: string
+  ownerRef?: string
   toolCall?: { toolName: string; args: unknown; toolCallId?: string }
 }
 
@@ -101,7 +103,7 @@ async function loadPersona(): Promise<string> {
 async function runTool(
   toolName: string,
   args: unknown,
-  context: { prospectRef?: string } = {},
+  context: { ownerRef?: string; prospectRef?: string } = {},
 ): Promise<{ ok: boolean; output: unknown }> {
   try {
     switch (toolName) {
@@ -166,6 +168,15 @@ async function runTool(
         const result = await buildCustomerBlueprintDraft(prospectRef)
         return { ok: result.ok, output: result }
       }
+      // FC-1A (#8637): authenticated owner-only coding fleet admission
+      // contract. The production route validates public-safe args but fails
+      // closed until a real durable fleet-run authority is injected.
+      case "coding_fleet_start": {
+        const result = await startSarahCodingFleetRun(args, {
+          ownerRef: context.ownerRef,
+        })
+        return { ok: result.ok, output: result }
+      }
       default:
         return {
           ok: false,
@@ -198,6 +209,7 @@ export async function runOwnedSarahTurn(
     toolCallId = `tool.${crypto.randomUUID()}`,
   ) => {
     const result = await runTool(toolName, args, {
+      ...(input.ownerRef ? { ownerRef: input.ownerRef } : {}),
       ...(input.prospectRef ? { prospectRef: input.prospectRef } : {}),
     })
     const recorded = {
@@ -381,4 +393,5 @@ export const SARAH_OWNED_TOOL_INVENTORY = [
   "live_stats",
   "plan_catalog",
   "customer_blueprint_draft",
+  "coding_fleet_start",
 ] as const
--- a/apps/sarah/src/server.test.ts
+++ b/apps/sarah/src/server.test.ts
@@ -1,8 +1,13 @@
-import { describe, expect, test } from "bun:test"
-import { unlink } from "node:fs/promises"
+import { afterEach, describe, expect, test } from "bun:test"
+import { mkdir, readFile, rm, unlink, writeFile } from "node:fs/promises"
 import { join } from "node:path"
 import { handleSarahRequest } from "./server.ts"
 import { SARAH_OWNED_TOOL_INVENTORY } from "./agent-runtime/owned-runtime.ts"
+import {
+  __setSarahCodingFleetRunStoreForTest,
+  createSarahCodingFleetFileStoreForTest,
+  listSarahCodingFleetRunsForTest,
+} from "./services/coding-fleet.ts"
 import {
   __resetCustomerBlueprintForTest,
   __setCustomerBlueprintLatestDraftReaderForTest,
@@ -10,7 +15,19 @@ import {
   CUSTOMER_BLUEPRINT_SCHEMA,
 } from "./services/customer-blueprint.ts"
 
+function installCodingFleetFileStoreForTest(fleetFile: string) {
+  const fleetPath = join(process.cwd(), ".sarah", fleetFile)
+  __setSarahCodingFleetRunStoreForTest(
+    createSarahCodingFleetFileStoreForTest(fleetFile),
+  )
+  return fleetPath
+}
+
 describe("apps/sarah monorepo service", () => {
+  afterEach(() => {
+    __setSarahCodingFleetRunStoreForTest(null)
+  })
+
   test("ops endpoint describes /sarah mount and rails", async () => {
     const res = await handleSarahRequest(
       new Request("http://localhost/sarah/api/operator/ops"),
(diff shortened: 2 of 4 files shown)


AtlantisPleb (changes requested):
Read-only integration review against current main 862fbefa88:\n\nGreen:\n- merge-tree is conflict-free;\n- after installing the merged lockfile, bun test apps/sarah/src/server.test.ts -t coding_fleet_start passes 3/3;\n- bun run --cwd apps/sarah typecheck passes;\n- the scope is appropriately narrow and does not add executor authority.\n\nChanges required before this can honestly call the pending request durable:\n\n1. readIndex silently converts any unreadable, truncated, or schema-mismatched existing file into an empty index. Because writeIndex writes directly to the authority file, a crash or partial write can make the next request erase the prior run registry. Existing invalid state must fail closed and preserve the evidence; writes need an atomic temp-file plus rename boundary (or, preferably, the existing typed Pylon/server durable store).\n\n2. writeQueue = writeQueue.then(...) stays rejected forever after one filesystem failure. Every later fleet start skips its callback and fails. Recover the serialization tail without swallowing the current request error, and test a failed write followed by a successful request.\n\n3. The repository contract requires new TypeScript boundaries to use Effect and Effect Schema. This file currently hand-parses the request and casts the JSON store. Please define/decode the request/index/record with Effect Schema, including rejecting duplicate plan unit refs and invalid dependencies/cycles rather than persisting an unusable DAG.\n\nAlso please add the named two-owner isolation test and reconsider returning ownerHash publicly: it is a stable cross-run owner fingerprint and is not needed to subscribe by runRef/scope.\n\nThis remains a useful FC-1A slice after these corrections; it still will not close #8637, whose relationshipMode, server-authority handoff, latency budgets, first claim, and E2E remain explicit residual.

trigger-xyz:
Pushed review fixes in `c81fbb7248`.

Changes made against the requested items:

- `readIndex` now treats only a missing store as empty. Unreadable, truncated, invalid JSON, or schema-mismatched existing stores fail closed with `store_unavailable` and do not overwrite the evidence.
- `writeIndex` now writes a same-directory temp file and `rename`s it into place atomically.
- `writeQueue` now recovers its serialization tail after failures while still returning the current request's store error. Added a failed-write-then-successful-request test.
- Added Effect Schema decoding for the tool request and persisted run index/record, reusing the shared `FleetWorkerKind` schema from `@openagentsinc/khala-fleet-intents`.
- Added plan DAG validation for duplicate unit refs, unknown dependency refs, and dependency cycles.
- Added the two-owner isolation test with the same idempotency key.
- Removed `ownerHash` from public tool output.

Validation:

- `bun test apps/sarah/src/server.test.ts -t coding_fleet_start` — 7 pass
- `bun test apps/sarah/src/server.test.ts` — 33 pass
- `bun run --cwd apps/sarah typecheck`
- `bun run --cwd apps/sarah test` — 274 pass
- `git diff --check`

AtlantisPleb (changes requested):
Second read-only review of c81fbb7248 against current main 981b359196.

The first requested fixes are present: invalid existing JSON fails closed, writes use same-directory temp+rename, the queue tail recovers, Effect Schema decodes request/store shapes, DAG duplicates/unknown deps/cycles are refused, the two-owner test exists, and ownerHash is gone.

Three correctness blockers remain:

1. This is not yet a durable server authority in the production topology. The default store is .sarah/coding-fleet-runs.json under process.cwd(), protected only by one process-local promise queue. A container/host replacement loses it, multiple Sarah replicas can race read-modify-rename and drop each other's runs, and Pylon has no authoritative claim path from it. Yet coding_fleet_start is added to the active inventory and returns ok:true / pending_executor. Please either persist through the existing owner-scoped server authority that Pylon can claim, or keep this explicitly fixture-only/default-off and return a typed executor/store-unavailable result in production. Do not expose a successful "durable pending request" backed only by an ephemeral local JSON file.

2. Public error output can echo the private input and local filesystem. decodeRequest returns the Effect parse message verbatim (bounded but unsanitized). Reproduction with the same Effect decoder:
   Expected string, got {"secret":"OPENAGENTS_AGENT_TOKEN=abc"}
The store error path likewise embeds Node read/write errors, which normally include the absolute index path. The tool result is a Sarah/public projection boundary. Return typed field/code plus fixed safe copy; keep raw schema/fs diagnostics owner-local.

3. Store decoding checks shape, not the same public-safe refinements used at write time or index consistency. A schema-valid tampered/stale record can carry a local path or credential-shaped objective/verifier and publicResult returns it on idempotent replay. Validate every loaded record through the canonical safe schemas/refinements and fail closed if run key/runRef, owner-scoped idempotency mapping, scope, or public text invariants disagree.

Integration note: the PR is currently conflict-dirty only on the additive @openagentsinc/khala-sync dependency in apps/sarah/package.json and bun.lock; rebase/retain both dependencies. I could not run a current-main merged gate until those conflicts are resolved.

This can remain FC-1A with relationshipMode, first claim/latency, controls, and E2E explicitly residual, but the intermediate success path itself must be honest and production-safe.

trigger-xyz:
Second review pass addressed in `90107d69c9`.

Changes:
- Rebased onto current `origin/main` (`5008856f57`) and retained both Sarah deps from the rebase conflict (`khala-fleet-intents` and `khala-sync`).
- Kept the local JSON fleet-run store fixture-only/default-off behind `SARAH_CODING_FLEET_START_LOCAL_STORE=1`; valid authenticated production calls now return fixed `store_unavailable` instead of claiming durable pending state.
- Replaced raw Effect schema/store error echoes with fixed public-safe errors, so malformed args and store failures cannot leak private input or local paths.
- Added loaded-store invariant validation before replay: run key/runRef, owner-scoped idempotency mapping, scope, canonical public text, verifier, repository, work source, worker policy, and target concurrency now fail closed if tampered.

Validation:
- `bun run --cwd apps/sarah typecheck`
- `bun test apps/sarah/src/server.test.ts` (36 pass, 0 fail)
- `bun run --cwd apps/sarah test` (352 pass, 0 fail)
- `git diff --check`
- `git diff --cached --check`

Intentionally still not changed:
- No real production durable Pylon/backend start path is claimed here; the PR remains a safe Sarah-side guard/fixture layer until a durable fleet-run authority is wired.

AtlantisPleb (changes requested):
Third read-only review of 90107d69c9, merged locally onto current main c742e6e73b.\n\nThe previous three blockers are materially addressed: production-default calls return fixed store_unavailable; schema/filesystem diagnostics no longer reach the tool result; loaded records and owner-scoped mappings are revalidated; the current-main merge passes 36 server tests and Sarah typecheck.\n\nTwo merge blockers remain for this partial guard/fixture slice:\n\n1. The local JSON path is described as fixture-only, but it is still a production-route runtime feature switch. Any deployed process with SARAH_CODING_FLEET_START_LOCAL_STORE=1 lets an authenticated /sarah tool call return ok:true/pending_executor from process-local JSON, recreating the exact false durable-authority state this revision is meant to prevent. Make the JSON store an injected test adapter that the production server/owned-runtime call site cannot select, or otherwise make the live route unconditionally store_unavailable until a real durable authority is injected. Update the owned-runtime comment too: it currently says this case creates a durable pending request even when the production path intentionally cannot.\n\n2. The fixture store/idempotency boundary still has unsafe ambiguity. SARAH_CODING_FLEET_RUNS_PATH is joined without rejecting absolute or parent traversal, so an enabled fixture can escape .sarah and atomically replace another local file. Also, reusing one owner/idempotency key with different objective/repository/verifier/work source silently replays the first run; store a canonical request fingerprint and return a fixed idempotency_conflict instead. Frame the run digest input unambiguously (canonical JSON or length prefixes) and validate the bounded owner ref before using it as a map/hash input.\n\nAfter those changes this looks mergeable as an explicitly non-production FC-1A admission contract and fixture, not as #8637 completion. The production durable Sarah authority, Pylon claim path, relationshipMode policy, latency gates, and integrated C1 proof remain residual.

trigger-xyz:
Third review pass addressed in `37735f0657`.

Changes:
- Rebased onto current `origin/main` (`418878a11e`).
- Removed the production-route env switch for the local JSON store. `SARAH_CODING_FLEET_START_LOCAL_STORE` / `SARAH_CODING_FLEET_RUNS_PATH` no longer enable `ok:true` from the live Sarah route; without an injected store the route returns fixed `store_unavailable`.
- Moved the JSON store behind an explicit test adapter (`createSarahCodingFleetFileStoreForTest` + test-only injection) so the owned-runtime/server call site cannot select it by process env.
- Updated the owned-runtime comment to stop claiming a durable pending request on the production path.
- Rejected fixture store paths that are absolute or contain parent traversal before filesystem access.
- Added canonical JSON hashing for owner/idempotency run refs and stored request fingerprints.
- Added `idempotency_conflict` when the same owner/idempotency key is reused with a different normalized request instead of silently replaying the first run.
- Loaded fixture records now also validate the stored request fingerprint against canonical public-safe fields.

Validation:
- `bun run --cwd apps/sarah typecheck`
- `bun test apps/sarah/src/server.test.ts -t coding_fleet_start` (11 pass, 0 fail)
- `bun test apps/sarah/src/server.test.ts` (37 pass, 0 fail)
- `bun run --cwd apps/sarah test` (414 pass, 0 fail)
- `git diff --check`
- `git diff --cached --check`

Still intentionally residual:
- This remains an FC-1A admission contract and fixture slice only. Production durable Sarah authority, Pylon claim path, relationshipMode policy, latency gates, and integrated C1 proof remain outside this PR.

AtlantisPleb (approved):
Fourth review of 37735f0657 on a clean merge with current main 9d026d313c: the two remaining blockers are closed. The JSON authority is now test-injected only and cannot be selected by the production route; fixture paths reject absolute/parent escape; canonical request fingerprints produce fixed idempotency_conflict; loaded fixture records are revalidated; errors stay fixed/public-safe. Canonical verification: 11 focused coding_fleet_start tests passed, full Sarah suite 414 passed, Sarah typecheck passed, and diff check passed. Approved strictly as FC-1A admission/fixture guard. It does not provide production durability, relationship policy, Pylon claim/activation, latency proof, or C1 completion.

Issue #8712, opened 2026-07-11
EP250 'Ready the Fleet': Desktop fleet overview, harness-selectable new chat, Fable (Claude) streaming


AtlantisPleb:
Owner-directed pull-forward for the Episode 250 recording ("Ready the Fleet"). This is a bounded slice under program #8566, pulling forward narrow sub-scopes of CUT-25 (#8705), CUT-21 (#8701), and CUT-16 (#8696) without claiming their exits. Coordinator: Fable (Claude session), working from a fresh clean worktree off `origin/main`.

## Compressed plan

**Lane A — Fleet overview workspace (sub-scope of #8705, read-only overview, not the authoritative cockpit exit):**
- New `fleet` workspace in `apps/openagents-desktop` with a new button at the top of the left sidebar dock (`shell.ts` NavRail), plus a command-palette entry.
- Overview shows Arbiter-discipline status dots — a dot lights only from decoded fresh evidence (account readiness from `pylon accounts list --json`, usage truth from `pylon accounts usage --refresh --json`); no receipt, no light — plus a standard table (provider, account ref, email, readiness, usage where fetched).
- Management stays in Settings: fleet view buttons deep-link via existing typed intents (`DesktopSettingsToggled`); no account mutation from the fleet page.
- New host adapter `provider-accounts` following the shipped `codex-connect.ts` pattern (spawn Pylon CLI read-only, decode typed projection, dedicated IPC channels). No Runtime Gateway protocol change for fleet data.

**Lane B — Harness-selectable new chat + Fable streaming (sub-scopes of #8696/#8701):**
- Composer/new-chat harness selector: `Fable` (lane `claude_pylon`) | `Codex` (lane `codex_app_server`).
- Thread the choice renderer → gateway `conversation.start` as an **optional additive field** (no protocol version bump; existing peers unaffected) → replace the hardcoded `target: { lane: "codex_app_server" }` in `main.ts` with the selected lane.
- Settings gains a Claude accounts section (list + readiness via the same provider-accounts projection; connect flow remains Codex-only in this slice).
- The downstream path already exists and is not modified here: Khala Sync `runtime.startTurn` → Pylon runtime-intent supervisor → `claude-agent-executor` → `openagents.khala_runtime_event.v1` events → confirmed timeline → provider-agnostic renderer. Protected Pylon fleet core (`apps/pylon/src/orchestration`, `src/node`) is not touched.

[...]

PR #8713, opened 2026-07-11
fix(desktop): fail closed for local Fable sends

trigger-xyz:
## Problem

Part III of the #8712 after-action identified an inert affordance: when Desktop falls back to local chat mode, selecting Fable can still press Send, but the route ignores the selector and falls through to the legacy local gateway path.

## Change

- Wrap the local chat host selected by `selectDesktopChatHost` with a fail-closed Fable guard.
- Return a clear local-mode error for `harness: "fable"` before the legacy local send host is called.
- Keep runtime Sync-mode lane routing unchanged: live `conversation.start` still maps Fable to `claude_pylon` and Codex to `codex_app_server`.
- Extend the runtime-conversation test to cover local unavailable/catching-up modes, proving Fable does not silently substitute while Codex/local sends still reach the existing host.

## Files

- `apps/openagents-desktop/src/renderer/runtime-conversation.ts`
- `apps/openagents-desktop/src/renderer/runtime-conversation.test.ts`

## Validation

- `bun test apps/openagents-desktop/src/renderer/runtime-conversation.test.ts` — 8 pass / 0 fail
- `bun run --cwd apps/openagents-desktop typecheck` — pass
- `bun run --cwd apps/openagents-desktop test` — 239 pass / 0 fail / 1281 expects (run outside sandbox because the PKCE tests bind a loopback listener)

## Value

This closes one mode x lane x action cell from the Part III failure: local mode x Fable selected x Send now fails with an explicit unavailable-lane message instead of hitting the dead legacy gateway and returning a model 400.

## Not Changed

- No local Claude streaming lane.
- No chip-disable UI.
- No legacy gateway model slug or endpoint repair.
- No CUT-10 runtime-live integration.
- No owner live-proof claim.

Refs #8712.

### DIFF
--- a/apps/openagents-desktop/src/renderer/runtime-conversation.test.ts
+++ b/apps/openagents-desktop/src/renderer/runtime-conversation.test.ts
@@ -20,11 +20,15 @@ describe("authoritative Runtime Gateway chat adapter", () => {
   })
 
   test("selects one mode at boot and retains local chat when Sync is unavailable", async () => {
+    const sent: Array<{ id: string; message: string; harness?: string }> = []
     const local: ChatHost = {
-      listThreads: async () => [],
+      listThreads: async () => [{ id: "thread.local.1", title: "Local", updatedAt: now, notes: [] }],
       newThread: async () => null,
       openThread: async () => null,
-      sendMessage: async () => ({ ok: false }),
+      sendMessage: async input => {
+        sent.push({ id: input.id, message: input.message, harness: input.harness })
+        return { ok: true }
+      },
     }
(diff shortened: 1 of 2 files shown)


trigger-xyz:
Closing as superseded by the stronger #8712 stack already landed on main: 765f1b5c69 adds the real local Fable lane, 899ba4ec9d registers the no-substitution behavior contract, 4436d7c09b adds the live-proof driver, and 7427ae9744 records live-proof receipts. That covers this PR's fail-closed/local-mode intent and more, so keeping #8713 open would only add a dirty duplicate.

PR #8722, opened 2026-07-12
fix(khala): split Pylon heartbeat diagnostics

### LINKED ISSUE #8716: Khala delegation gate: 409 `stale_or_missing_heartbeat` conflates aged-out vs never-heartbeated, and the resolver never echoes the selected pylonRef
### Affected surface

Khala coding delegation gate — `apps/openagents.com/workers/api/src/inference/coding-workflow-delegation.ts` (the `khala request --workflow codex_agent_task` admission path).

### Actual behavior

Two related diagnosability defects in the delegation admission gate, found while exercising the Khala → Pylon → Codex runbook on a host with more than one owner-linked Pylon:

1. **The `stale_or_missing_heartbeat` 409 conflates two very different states.** When an explicit `--pylon-ref` targets a Pylon whose server registration has genuinely lapsed, the caller gets `evidence.khala_coding.target_pylon_ref.unavailable.stale_or_missing_heartbeat` — the *same* evidence ref used when a Pylon never heartbeated at all. The reason string ("its online heartbeat is stale or missing") reads as "you never came online," which is alarming and wrong when the operator has other healthy Pylons and just pinned the wrong one. It sent me down a multi-attempt false-bug investigation.

2. **The caller-scoped resolver (no `--pylon-ref`) never reports which `pylonRef` it selected.** When the request omits `--pylon-ref`, the gate admits by routing to the freshest owner-linked Pylon and returns an `assignmentRef`, but nothing in the response says *which* Pylon was chosen. On a host with multiple owner Pylons, the operator cannot tell where work will run — so they cannot correlate an admitted assignment back to a specific Pylon/account, and cannot tell that "explicit-ref 409" and "resolver admits" refer to *different* Pylons.

Together these make a correct gate look like a freshness bug.

### Expected behavior

1. Split the diagnosis: a pinned Pylon that exists but whose heartbeat aged past TTL should return a distinct evidence ref/reason (e.g. `target_pylon_ref.heartbeat_stale` with "this specific pinned Pylon last heartbeated <n>s ago; another of your linked Pylons may be available — omit `--pylon-ref` to use caller-scoped selection") vs. a Pylon with no heartbeat ever (`heartbeat_missing`).

2. The admission response (both explicit-ref and resolver paths) should echo the **selected `pylonRef`** so the operator can correlate the admitted assignment to a specific Pylon.

### Reproduction steps

On a host with two or more owner-linked Pylons where at least one is heartbeat-fresh and one is stale:

1. `pylon presence heartbeat` for Pylon A (fresh); leave Pylon B's server registration stale.
2. `pylon khala request --workflow codex_agent_task --pylon-ref <B> --no-run …` → **409** `stale_or_missing_heartbeat` (correct that B is stale, but the ref/reason implies "never online").
3. Same instant, `pylon khala request --workflow codex_agent_task --no-run …` (no `--pylon-ref`) → **admits**, returns an `assignmentRef`, but does not say it selected A.
4. Observe: no way from the two responses to tell they evaluated different Pylons.

`--no-run` gate probes reproduce this without consuming any execution account.

### Public-safe evidence

- Gate freshness check: `hasFreshOnlineHeartbeat` at `coding-workflow-delegation.ts` requires `now - latestHeartbeatAt <= 5min` AND status ∈ {available,healthy,idle,online,ready}; the failure is reported through the single `stale_or_missing_heartbeat` sub-condition for both the aged-out and never-heartbeated cases.
- Registrations are one-row-per `pylon_ref` (upsert `ON CONFLICT(pylon_ref) DO UPDATE SET latest_heartbeat_at`), so explicit-ref and resolver read the same freshness data; the divergence is purely *which Pylon* each path evaluates.
- Full diagnosis and repro receipts: OpenAgents issue #8712 (comments dated 2026-07-12, "Lane RE-B RE-DIAGNOSED").

### Severity

S2 - degraded, confusing, stale, or partially broken behavior

### Environment

- Repo: `OpenAgentsInc/openagents`, `main` at/after `ce9a0a422e` (2026-07-12).
- Local Pylon CLI (`bun apps/pylon/src/index.ts`) against the live `openagents.com` Worker delegation gate.
- Reproduced with `PYLON_DISABLE_DAEMON_ROUTING=1` and `--no-run` gate probes; multiple owner-linked Pylons on one host.

### Safety and redaction

No secrets, tokens, wallet material, private prompts, or provider payloads included. Pylon refs are opaque public identifiers; account/credential material is not referenced. This is a diagnosability/observability bug in a public gate, not a spend/authority/privacy issue.

### Notes

Explicitly out of scope (separately known, not this issue): account-pool depletion pre-dispatch visibility, and transient `workspace_dependency_install_failed` / `codex_custody.network` failures under heavy parallel load.

trigger-xyz:
## Problem

Fixes #8716.

The Khala coding delegation gate returned the same `stale_or_missing_heartbeat` diagnosis for two different explicit target-Pylon states: a Pylon that never heartbeated and a Pylon whose heartbeat aged past the freshness TTL. That made a correct multi-Pylon admission path look like a freshness bug. The issue also asked that admitted responses make the selected Pylon clear for resolver-selected capacity.

Claim: https://github.com/OpenAgentsInc/openagents/issues/8716#issuecomment-4950615855

## Change

- Splits targeted-Pylon heartbeat diagnosis into explicit subconditions:
  - `heartbeat_missing`
  - `heartbeat_stale`
  - `heartbeat_status_unavailable`
- Makes stale-heartbeat refusal text name the pinned-Pylon age and the 300s TTL, and directs operators to omit `--pylon-ref` for caller-scoped selection when another linked Pylon may be available.
- Adds regression coverage that the chat-completions delegation response includes selected `pylonRef` metadata in the SSE payload, alongside the existing selected-Pylon header.
- Keeps GEPA feedback parsing compatible with historical `stale_or_missing_heartbeat` text while recognizing the new refs.

## Validation

- `bun run --cwd apps/openagents.com/workers/api vitest run src/inference/coding-workflow-delegation.test.ts src/inference/chat-completions-routes.test.ts src/khala-delegation-gepa-feedback.test.ts`
  - 3 files passed, 220 tests passed.
- `NODE_OPTIONS=--max-old-space-size=8192 bun run --cwd apps/openagents.com/workers/api typecheck`
  - Passed.

## Not Changed

No dispatch authority widening, no Pylon registration schema/migration, no product-promise/OpenAPI versioning, no spend, no account-pool depletion fix, and no workspace dependency/install failure handling.

### DIFF
--- a/apps/openagents.com/workers/api/src/inference/chat-completions-routes.test.ts
+++ b/apps/openagents.com/workers/api/src/inference/chat-completions-routes.test.ts
@@ -479,6 +479,9 @@ describe('coding delegation default-on guard', () => {
     expect(response.headers.get('openagents-selected-pylon-ref')).toBe(
       'pylon.owner.codex',
     )
+    const text = await response.text()
+    expect(text).toContain('"pylonRef":"pylon.owner.codex"')
+    expect(text).toContain('"workflowClass":"codex_agent_task"')
   })
 
   test('direct agent-owned Pylon dispatch survives a transient OpenAuth link read failure', async () => {
--- a/apps/openagents.com/workers/api/src/inference/coding-workflow-delegation.test.ts
+++ b/apps/openagents.com/workers/api/src/inference/coding-workflow-delegation.test.ts
@@ -958,11 +958,54 @@ describe('coding workflow delegation', () => {
     expect(result).toMatchObject({
       error: 'target_pylon_unavailable',
       evidenceRefs: expect.arrayContaining([
-        'evidence.khala_coding.target_pylon_ref.unavailable.stale_or_missing_heartbeat',
+        'evidence.khala_coding.target_pylon_ref.unavailable.heartbeat_stale',
       ]),
       kind: 'rejected',
       statusCode: 409,
     })
+    if (result?.kind !== 'rejected') {
+      throw new Error('expected a rejection')
+    }
+    expect(result.reason).toContain('last heartbeated 3600s ago')
+    expect(result.reason).toContain('omit --pylon-ref')
+  })
+
+  test('diagnoses a never-heartbeated target separately from a stale target (#8716)', async () => {
+    const result = await delegateCodingWorkflow({
+      classification,
+      linkedAgents: [linkedOwner],
+      makeId: () => 'id1',
+      nowIso,
+      pylonStore: makeStore({
+        registrations: [
+          registration({
+            latestHeartbeatAt: null,
+            latestHeartbeatStatus: null,
+          }),
+        ],
+      }),
+      rawBody: {
+        openagents: { coding: { targetPylonRef: 'pylon.owner.codex' } },
+      },
+      requestId: 'chatcmpl_coding_missing_heartbeat',
+    })
+
+    expect(result).toMatchObject({
+      error: 'target_pylon_unavailable',
+      evidenceRefs: expect.arrayContaining([
+        'evidence.khala_coding.target_pylon_ref.unavailable.heartbeat_missing',
+      ]),
+      kind: 'rejected',
+      statusCode: 409,
+    })
+    if (result?.kind !== 'rejected') {
+      throw new Error('expected a rejection')
+    }
+    expect(result.evidenceRefs).not.toContain(
+      'evidence.khala_coding.target_pylon_ref.unavailable.heartbeat_stale',
+    )
+    expect(result.reason).toContain('no recorded online heartbeat yet')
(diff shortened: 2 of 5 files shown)


AtlantisPleb (approved):
Reviewed against current origin/main. The heartbeat diagnosis split preserves admission policy and authority, the selected-pylon response metadata is regression-covered, the patch merges cleanly, 220 focused tests pass, and the full API typecheck passes.

